PR #8, opened 2015-10-26
Support IIIF Authentication for requesting resources

cbeer:
In f8125401e808b5d0090972692dae42af4424ce60, we added support for generating IIIF Authentication tokens, but were still relying on the login service to maintain a WebAuthed session. b2c8dbb finishes that work by supporting Bearer token authentication in addition to WebAuth + Basic Auth.

### DIFF
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -1,18 +1,18 @@
 ##
 # API for delivering IIIF-compatible images and image tiles
 class IiifController < ApplicationController
-  before_action :load_image, except: [:token]
-  before_action :add_iiif_profile_header, except: [:token]
+  before_action :load_image
+  before_action :add_iiif_profile_header
 
   rescue_from ActionController::MissingFile do
     render text: 'File not found', status: :not_found
   end
 
-  before_action only: :show do
+  before_action except: :metadata do
     fail ActionController::MissingFile, 'File Not Found' unless @image.valid?
   end
 
-  before_action only: :meadata do
+  before_action only: :metadata do
     fail ActionController::MissingFile, 'File Not Found' unless @image.exist?
   end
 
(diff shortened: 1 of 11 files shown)


mejackreed:
Not sure if this is relevant here, but in embed we cannot completely adhere to the IIIF with spec. Because of our dependencies we aren't at this time able to include `access_token` on every request to stacks. Is this going to be an issue?

cbeer:
Can you say more?

In practice, it shouldn't matter too much:
- Requests for resources will either work / not work based on the user's webauth status
- Requests for info.json have minor differences for the different scenarios. E.g. for an image only Stanford users can download, when you are not auth'ed, the info.json will only contain a pre-cast size for the `!400x400` thumbnail. For Stanford auth'ed users, it will have the various sizes (100%, 50%, 25%, etc). It would be nice if there was a way to express this access control in the info.json file itself (@azaroth42?) so the resource could be cached for all users.
- Requests for manifest.json resources are identical under all scenarios

mejackreed:
:ok: this seems to be ok. To your first point, that's great, I think there isn't an issue then.

We will not be able to meet this *must* threshold:
> Once obtained, the token value must be passed back to the server on all future requests via XMLHttpRequests by adding an Authorization request header, with the value Bearer TOKEN_HERE. The token should be added to all requests for resources from the same domain and subdomains that have a reference to the service, regardless of which API is being interacted with.

http://iiif.io/api/auth/0.9/#json-interaction

So for some info.json requests, we will not be able to send the `accessToken`, even after we are authenticated. This is due to how our dependencies (manifestor and OpenSeadragon) call info.json resources.

I **think** we ok.

cbeer:
:ok_hand: ideally, IIIF would give us some way to express that a little better, but I agree, it's probably no big deal.

app/controllers/iiif_controller.rb
```diff
@@ -38,19 +38,30 @@ def metadata
...
+                       accessToken: encode_credentials(current_user.token).sub('Bearer ', ''),
+                       tokenType: 'Bearer',
+                       expiresIn: 3600
+                     }
                    else
                      { error: 'missingCredentials', description: '' }
                    end
 
+        if response[:accessToken] && !current_user.webauth_user?
+          cookies[:bearer_token] = { value: response[:accessToken], expires: 1.hour.from_now, httponly: true }
```
cbeer: Check protocol and create a secure cookie if HTTPS.

PR #15, opened 2015-11-03
Provide scaleFactors that are powers of 2

### LINKED ISSUE #14: change scaleFactors to powers of 2
In the current deployed version, the scaleFactors are, for example, `[1,2,3,4,5]` for a 17000 pixel wide image.  This means that tiles are available at 1/1, 1/2, 1/3, 1/4 and 1/5 scales.

Open Seadragon will only request factors of 2: 1/1, 1/2, 1/4, 1/8, 1/16 and so forth, and thus only 1/1, 1/2 and 1/4 are ever requested from the scaleFactors provided.

The fix (IMO) is to return scaleFactors that are powers of 2, such that the long edge of the largest scaleFactor is one step below the tilesize, thus the most zoomed out will be four tiles.

mejackreed:
Closes #14 . Should also help with https://github.com/sul-dlss/sul-embed/issues/484 , by providing a more performant tileresponse

I'm a bit weary though, this doesn't feel like the right way to do this. Should we be changing the calculation here:

https://github.com/jronallo/djatoka/blob/e8d4f77311c47563da35f46a79c2508cc215593f/lib/djatoka/metadata.rb#L60-L77

or here:

https://github.com/jronallo/djatoka/blob/e8d4f77311c47563da35f46a79c2508cc215593f/lib/djatoka/metadata.rb#L159-L161

Do the `scaleFactors` correspond to the calculated sizes at those levels?

### DIFF
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -68,7 +68,7 @@ class IiifController < ApplicationController
     @image ||= StacksImage.new(image_params)
   end
 
-  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
   def image_info
     info = @image.info do |md|
       if can? :download, @image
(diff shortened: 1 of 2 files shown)


mejackreed:
Also cc @azaroth42 ...

mejackreed:
Also this doesn't address the fix in the way that @azaroth42 described. Waiting to discuss with @cbeer first.

cbeer:
I don't know anything. If @azaroth42 says it's what we should do, should we change it in the djatoka gem?

cbeer:
Fixed by b5190340a183b831da78b8bf351e1e4dc1a2b2dc

PR #26, opened 2016-03-01
remove identifier from downloaded filename

peetucket:
per conversation with Stu and @azaroth42 , for better user experience

### DIFF
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -71,7 +71,7 @@ class IiifController < ApplicationController
   end
 
   def set_attachment_content_disposition_header
-    response.headers['Content-Disposition'] = "attachment;filename=#{params[:identifier]}.#{params[:format]}"
+    response.headers['Content-Disposition'] = "attachment;filename=#{identifier_params[:file_name]}.#{params[:format]}"
   end
 
   def load_image
--- a/spec/controllers/iiif_controller_spec.rb
+++ b/spec/controllers/iiif_controller_spec.rb
@@ -78,7 +78,7 @@ describe IiifController, :vcr do
       end
 
       it 'sets the preferred filename' do
-        expect(subject.headers['Content-Disposition']).to include 'filename=nr349ct7889%2Fnr349ct7889_00_0001.jpg'
+        expect(subject.headers['Content-Disposition']).to include 'filename=nr349ct7889_00_0001.jpg'
       end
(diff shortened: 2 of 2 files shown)


ndushay:
:+1:  from me, but I'd be more comfy if someone with digital_stacks_rails knowledge would weigh in.  @cbeer - would that be you?

app/controllers/iiif_controller.rb
```diff
@@ -71,7 +71,7 @@ def set_image_response_headers
   end
 
   def set_attachment_content_disposition_header
-    response.headers['Content-Disposition'] = "attachment;filename=#{params[:identifier]}.#{params[:format]}"
+    response.headers['Content-Disposition'] = "attachment;filename=#{identifier_params[:file_name]}.#{params[:format]}"
```
jkeck: Hmm, okay, so the only question I have is are are we 100% sure all the identifiers will always include the slash (encoded as it is now)?  If it does not included the slash, we'll pass `nil` here (which I'm not exactly sure what would be the end-user experience there).

That could be a totally baseless concern, so I'm fine with :ship:ing and fixing if we find that we do have identifiers that do not include the `/`.

peetucket: My understanding was the identifier always included "druid/" in our case, so it **should** always work.  I wasn't able to identify a more direct way of fetching the filename from the params without going through the identifier_params method (which does rely on splitting on the encoded / character).  Can you see a way of avoiding this roundabout method of getting the filename?

cbeer: Wait, we're assuming the druid is already present in the file name? That's certainly not the case with all of our image objects (see https://purl.stanford.edu/pt257cr7719, but there's even worse cases like "00001.jp2" floating around too )

peetucket: No, definitely not - filenames do not usually include the druid.  The goal is to have the downloaded filename be just the filename as it actually exists in the object and without the pre-prepended object identifier (currently DRUID/).   This was the original behavior before IIIF was implemented.  I have confirmed with Rob and Stu that it is acceptable and within IIIF spec to go back to this behavior.  The discussion is about the best way to re-implement this.

peetucket:
@LynnMcRae gave this functionality change a :+1: this morning

jkeck:
Will back this out later if it blows everything up. :boom:

Issue #29, opened 2016-04-14
Media stacks - Evaluate manifest/playlist redirects in stacks


peetucket:
Evaluate how to redirect playlist M3U8 and manifest (MPD) to wowza server from stacks URL based on user's browser

we need to support two types of manifest files since varying client support requires it

Also:  Evaluate whether difference between HLS and MPEG-dash is important 

HLS and MPEG-dash both make HTTP requests for the manifest, but

HLS makes HTTP range queries to get it in chunks

MPEG-dash has the client make TCP connections to get data

LynnMcRae:
This needs to be part of the complete specification of the stacks/media URL syntax. So far this is the only addition to the URL beyond identifying the object and resource desired but we don't know if that's all we need

ndushay:
We think this is Rails level work, rather than Apache level work.

peetucket:
possibly related to/duplicates #29

ndushay:
#33 is closed;  leaving this open until we have dash format working in the tracer bullet.

PR #38, opened 2016-04-26
implement media in stacks

ndushay:
- allows for download or streaming of media files
- set up to work for mp3, mp4 "containers";  mp3, mp4, mov, mpeg  file extensions.

note:   this PR does not include authorization with wowza, but does include dor-rights-auth.

connects to #29


app/controllers/media_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    if current_user
+      super(exception)
+    elsif params['action'] == 'stream'
+      redirect_to auth_media_stream_url(params.symbolize_keys)
+    else
+      redirect_to auth_media_download_url(params.symbolize_keys)
+    end
+  end
+
+  def media_params
```
jkeck: Possibly use strong params here? http://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters

Particularly when we `redirect_to` params above.

app/controllers/media_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+# API for delivering streaming media via stacks
+class MediaController < ApplicationController
+  before_action :load_media
+
+  rescue_from ActionController::MissingFile do
+    render text: 'File not found', status: :not_found
+  end
+
+  def download
+    return unless stale?(cache_headers)
```
jkeck: This seems to be copied over from file.  Does it make sense here?  Should we remove it until we have a reason to re-implement?

app/models/stacks_media_stream.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def streaming_file_prefix
+    if video?
+      'mp4'
+    elsif audio?
+      'mp3'
+    end
+  end
+
+  def streaming_url_for(type)
```
jkeck: Consider placing this method at the top of the private section, since it is the main private interface of the public interface.

config/settings.yml
```diff
@@ -2,6 +2,20 @@ stacks:
   storage_root: /stacks
   djatoka_url: 'http://imageserver-b.stanford.edu/adore-djatoka/resolver'
 
+purl:
+  url: 'https://purl.stanford.edu/'
+
+stream:
+  url: http://streaming-server.com:6666/stuff
```
jkeck: Can we put the default wowza port here so we don't have to guess?

config/settings/development.yml
```diff
@@ -0,0 +1,2 @@
+app_users:
```
jkeck: Can we validate if we need this in our `development.yml` file?

spec/controllers/media_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    subject { get :download, id: 'bb582xs1304', file_name: 'bb582xs1304_sl', format: 'mp4' }
+
+    it 'sends the video url to the user' do
+      expect(controller).to receive(:send_file).with(video.path).and_call_original
+      subject
+    end
+
+    it 'loads the video' do
+      subject
+      video = controller.instance_variable_get(:@media)
```
jkeck: Is it possible to use `assigns` here instead of `instance_variable_get`?  If not, no worries, just a red-flag.

config/settings.yml
ndushay: will do.

app/models/stacks_media_stream.rb
ndushay: reordering the private methods.

spec/controllers/media_controller_spec.rb
ndushay: will change.  Note that instance_variable_get is used in spec/controllers/iiif_controller_spec.rb  and instance_variable_set is used in spec/controllers/file_controller_spec.rb

app/controllers/media_controller.rb
ndushay: removing it.

config/settings/development.yml
ndushay: don't need it.  w00t.

ndushay:
@jkeck  I implemented all the suggestions except I was having trouble with implementing strong parameters.   I tried a few things and it didn't work - I suspect there is some setting somewhere or a switch I don't know about that is in the way.   Feel free to go for it yourself, or we can work on it together tomorrow.

ndushay:
@jkeck and I just talked F2F:   we are creating a separate ticket for the strong params suggestion, so this PR is ok with the 2 of us as is.   Merge at will!

app/controllers/media_controller.rb
ndushay: We're going to do strong params as its own issue and PR, not here, per F2F w @jkeck

jmartin-sul: strong params seems like a good idea, but this is short and readable and seems to get the same thing done, so i'm also totally fine leaving it as-is.

spec/controllers/webauth_controller_spec.rb
```diff
@@ -30,6 +30,22 @@
...
+  describe '#login_media_download' do
+    subject { get :login_media_download, id: 'bb582xs1304', file_name: 'bb582xs1304_sl', format: 'mp4' }
+
+    it 'returns the user to the media download action' do
+      expect(subject).to redirect_to media_url(controller.params.symbolize_keys)
+    end
+  end
+
+  describe '#login_media_stream' do
+    subject { get :login_media_stream, id: 'bb582xs1304', file_name: 'bb582xs1304_sl.mp4', format: 'm3u8' }
```
jmartin-sul: totally possible that i'm missing something here, but would `format: 'mp4'` be more reasonable here?  (since it seems like `m3u8` isn't a media format).

app/models/stacks_media_stream.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def to_manifest_url
+    streaming_url_for(:dash)
+  end
+
+  private
+
+  def streaming_url_for(type)
+    return unless file_name && streaming_file_prefix
+    suffix = TYPE_TO_MANIFEST_FILE[type]
```
jmartin-sul: what about something like `playlist_filename` instead of `suffix`?

jmartin-sul:
two very minor questions/suggestions, but also happy to merge without implementing either.

spec/controllers/webauth_controller_spec.rb
ndushay: in this case, it is a "format" for the concept of the manifest for the container.  it is implying   playlist.m3u8   for hls, or  manifest.mpd for dash.

app/models/stacks_media_stream.rb
ndushay: manifest_filename  would be slightly more generic, imo, than "playlist".    it is being assigned from something called TYPE_TO_MANIFEST_FILE ...

PR #44, opened 2016-04-29
Strong params

### LINKED ISSUE #41: Implement strong params to avoid security risk
The worry is that "host" can be passed in as a param and thus the site could be easily spoofed.

ndushay:
closes #41 

(also added a commit to remove unused vcr cassettes;  can move to separate PR if preferred)


jmartin-sul:
:+1: as far as i can tell

app/controllers/iiif_controller.rb
```diff
@@ -108,8 +117,8 @@ def image_info
   end
   # rubocop:enable Metrics/MethodLength
 
-  def image_params
-    params.slice(:region, :size, :rotation, :quality, :format).merge(identifier_params).merge(canonical_params)
+  def stacks_image_params
+    allowed_params.permit(:region, :size, :rotation, :quality, :format).merge(identifier_params).merge(canonical_params)
```
jkeck: I'm curious if `permit` is the right thing to use here since `allowed_params` has already permitted those, so maybe `slice` is reasonable here?

app/controllers/media_controller.rb
```diff
@@ -29,21 +29,25 @@ def stream
 
   private
 
+  def allowed_params
+    params.permit(:action, :id, :file_name, :format)
```
jkeck: This is interesting, is `action` necessary to add here?  I don't recall needing to be explicit about things like `action` and `controller` before, but I could be totally wrong.

app/controllers/iiif_controller.rb
ndushay: good catch!

app/controllers/media_controller.rb
ndushay: when I take `action` out, I get 2 failures like this:

```
MediaController#stream for a restricted file with an unauthenticated user redirects to the webauth login endpoint for media stream
     Failure/Error: redirect_to auth_media_download_url(allowed_params.symbolize_keys)
     
ActionController::UrlGenerationError:
       No route matches {:action=>"login_media_download", :controller=>"webauth", :file_name=>"bb582xs1304_sl.mp4", :format=>"m3u8", :id=>"bb582xs1304"} missing required keys: [:file_name]
...
 # ./spec/controllers/media_controller_spec.rb:71:in `block (3 levels) in <top (required)>'
     # ./spec/controllers/media_controller_spec.rb:104:in `block (5 levels) in <top (required)>'
...
     # ./app/controllers/media_controller.rb:43:in `rescue_can_can'
...
     # --- Caused by: ---
     # CanCan::AccessDenied:
     #   You are not authorized to access this page.
     #   ./app/controllers/media_controller.rb:19:in `stream
```

app/controllers/file_controller.rb
```diff
@@ -17,18 +17,19 @@ def show
 
   private
 
+  def allowed_params
+    params.permit(:id, :file_name)
+  end
+  alias stacks_file_params allowed_params
+
```
tingulfsen: What is the purpose of adding this alias? It doesn't seem to add much.

tingulfsen:
Btw I am 👍  other than the minor comment above.

app/controllers/file_controller.rb
ndushay: My thought was that it was just a coincidence that the stacks_file_params and the allowed_params were the same.  They might change independently.  Perhaps this is a premature optimization, or perhaps it's a useful alias showing the differing functions of the two calls.

jmartin-sul:
can this be merged?  it appears that the potential remaining concerns are:
* specifically calling out `:action` in `MediaController#allowed_params`.
* the possibly superfluous `alias stacks_file_params allowed_params` in `FileController`.

sounds like the former might be necessary.  and the latter seems like it shouldn't be super-controversial regardless of whether it's left as-is?

app/controllers/file_controller.rb
jkeck: Perhaps we can add a comment here, just so a future developer will know that this is here for consistency, and for the ability to add a method that would winnow down the allowed params further (as is done in the other controllers).

I admit that at first the `alias` confused me until I saw the context of the rest of the PR.

app/controllers/media_controller.rb
jkeck: Hmm, I don't quite understand it, but that's okay. The tests indicate that there is a need somewhere.  I doubt this will have any negative consequences, so I'm 👍 to keeping it in unless a reason presents itself.

jkeck:
I ultimately think that the two issues shouldn't be blockers.  The change is 👍.

I have added comments about how we can move forward with (hopefully) minimal effort.

ndushay:
I added a comment about the alias to try to clarify it per @tingulfsen's question, and @jkeck gave 👍 to leaving the action in the allowed params.   So ... would someone like to press the big green button?

app/controllers/file_controller.rb
```diff
@@ -17,18 +17,20 @@ def show
 
   private
 
+  def allowed_params
+    params.permit(:id, :file_name)
+  end
+  # the args needed for StacksFile.new happen to be the same as allowed_params
+  alias stacks_file_params allowed_params
+
```
ndushay: Here is the comment added for the alias

jmartin-sul:
button pressed

PR #46, opened 2016-05-05
media token refactor:  token instead of token_string;  user_ip instead of user_ip_addr

ndushay:
I think "token" is good enough as a param name, but if we *must* have something more than that, I think "stacks_token" is better than "token_string".

Also not sure of underscores vs camel case in HTTP GET params.  Which is another reason why shorter feels better to me.

I also did a separate commit shortening user_ip_addr  to user_ip

@jmartin-sul

### DIFF
--- a/spec/controllers/media_controller_spec.rb
+++ b/spec/controllers/media_controller_spec.rb
@@ -68,17 +68,17 @@ describe MediaController, vcr: { record: :new_episodes } do
 
   describe '#stream' do
     let(:streaming_base_url) { Settings.stream.url }
-    let(:token_string) { 'tokenstring' }
-    let(:streaming_url_query_str) { "token_string=#{token_string}" }
+    let(:token) { 'encrypted_token_value' }
+    let(:streaming_url_query_str) { "token=#{token}" }
     subject { get :stream, id: 'bb582xs1304', file_name: 'bb582xs1304_sl.mp4', format: 'm3u8' }
 
     it 'redirects m3u8 format to streaming server mp4 prefix with playlist.m3u8' do
-      allow(controller).to receive(:media_token).and_return token_string
+      allow(controller).to receive(:media_token).and_return token
       streaming_url_path = '/bb/582/xs/1304/mp4:bb582xs1304_sl.mp4/playlist.m3u8'
       expect(subject).to redirect_to "#{streaming_base_url}#{streaming_url_path}?#{streaming_url_query_str}"
     end
     it 'redirects mpd format to streaming server mp4 prefix with manifest.mpd' do
-      allow(controller).to receive(:media_token).and_return token_string
+      allow(controller).to receive(:media_token).and_return token
       get :stream, id: 'bb582xs1304', file_name: 'bb582xs1304_sl.mp4', format: 'mpd'
       streaming_url_path = '/bb/582/xs/1304/mp4:bb582xs1304_sl.mp4/manifest.mpd'
       expect(response).to redirect_to "#{streaming_base_url}#{streaming_url_path}?#{streaming_url_query_str}"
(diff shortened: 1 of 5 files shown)


jkeck:
Hmm, I could go either way.  I'm partial to having something to identify that the token is from stacks so we're open to accepting tokens from other sources.

By being able to identify the tokens coming from stacks we won't have to modify the stacks code because we introduce another token coming from another source.

The would currently also have the benefit of allowing Wowza to infer WHERE to validate the token, although I'm not sure we've solved how we want to approach that yet, so it might be another mechanism entirely.

spec/controllers/media_controller_spec.rb
```diff
@@ -68,17 +68,17 @@
 
   describe '#stream' do
     let(:streaming_base_url) { Settings.stream.url }
-    let(:token_string) { 'tokenstring' }
-    let(:streaming_url_query_str) { "token_string=#{token_string}" }
+    let(:token) { 'encrypted_token_value' }
```
jmartin-sul: how about `encrypted_token` for this var name?  in other parts of the code (including tests below), `token` is a var name for a `StacksMediaToken` object and not its encrypted string representation.  seems like it'd be nice to be consistent.

ndushay: good idea.  will do.

ndushay:
We have two contexts going on here:
- stacks redirect:

```ruby
        redirect_to "#{@media.to_playlist_url}?token=#{media_token}"
```
In this spot, the param could be named `stacks_token=#{media_token}`  to be clear that the token is coming from the stacks.

- media token validation:

```
  (stacks_base_url)/media/id/filename.mp4/verify_token?token=encrypt_token&user_ip=192.168.1.100'
```
In this spot, it is clearly a token for stacks and also clearly a media token because of the url preceding.   "stacks_token" is redundant, and also would not necessarily distinguish it from other tokens.

I can understand the desire to have the same name parameter name in both places, but they are actually different contexts from the perspective of the requests being issued and received.

What am I missing?

ndushay:
I guess I could also argue for `authorize` as the action name instead of `verify_token` -- what do folks think about that?

jmartin-sul:
i like the idea of `stacks_token` as the URL param name.  the namespacing argument@jkeck makes seems sound to me.  though technically the iiif token is also a stacks token, since stacks is minting/reading it, right?  so i don't know... i'd vote for something like `stream_auth_token` or `media_token` if we wanted a better name than `token_string` (which was admittedly not my best naming moment).

i also really like the idea of generally using `token` as a var name to refer to a token (`StacksMediaToken`) object, and `encrypted_token` or some other word/phrase to refer to the token's encrypted string form.  since the one can't directly be substituted for the other without conversion, it'd be nice to have a terminology pattern that's easily skimmable, where the two aren't used interchangeably.

i think i'd prefer to not make URL params camel case.  languages where camel case is the usual style can declare their variables as such when reading/assigning the param value, but most of our work is in snake case, and i've always preferred the way that reads, so i'd prefer to use that in any situation where it doesn't go against prevailing norms.  though for HTTP headers, i'd def go with camel case, since that seems to be the norm there (i've seen plenty of both styles in GET and POST params).

totally fine dropping `_addr` from `user_ip`.

i dunno, i'd be :ok_hand: merging this, but it's about as readable for me as what was there before, though i find the new usage of the `token` name to be slightly less readable in light of some of the existing usage of that name.

FWIW, i'm probably less gung-ho about naming brevity than most people.  i like cryptic acronyms/abbreviations, and i like verbose descriptive phrases, but the area in between those extremes is my least favorite type of var name.  i like when a name tells me exactly what it is, or makes me look it up.

jmartin-sul:
re: `authorize` vs `verify_token`:  i'd be fine w/ that if it didn't step on cancan or rails terminology, but i'd avoid if either of those already used `authorize`.

ndushay:
`authorize` vs. `verify_token`:  you are correct, `authorize` is used by cancancancancan.   And `validate` is the ActiveModel validation thing.

Options:
- `verify_token` (as is)
- `validate` (would not be confused with ActiveModel validation because it would be an action method in the controller)
- ~~`authorize`~~ would be confusing with cancancan
- any other ideas?  

The reason I'm even considering a rename is because the spirit of the action is more about authorizing access, or validating the token, to my way of thinking.   It's not "yup, it's a token", but instead "yes, this token is valid for the id, filename and ip;  you may proceed"

jmartin-sul:
fair point.  to me, i don't know that `validate` feels closer to "yes, this token is valid for these values and is unexpired and was minted by stacks, you may proceed" than `verify_token`... and i personally prefer the latter just because it feels more descriptive as to what's happening.  and even if it doesn't conflict with ActiveModel's `validate` in terms of what gets executed, it feels like a somewhat confusing word reuse in a rails environment.  and `verify_token` lines up with the `StacksMediaToken` method name, though that wasn't a name i was super-happy with in the first place.

what about `token_grants_access`?  that's more explicit about the intended purpose of the method, but doesn't use rails or cancancancan terminology.

jmartin-sul:
if i had as many characters to spend as i wanted for a name, it'd be something like `verify_token_is_unexpired_for` or `verify_token_still_grants_access_to`, but those are kind of unwieldy.  but they are fully descriptive about all of the salient operations.

i think the reason i like "verify" more than "validate" is because validate sounds to me like checking that user input conforms to what a given field allows or what another piece of data constrains it to.  "verify" sounds to me more like "was this issued by the entity that claims to have issued it?"  signalling the cryptographic verification seems more important than the input validation, since the latter is necessarily implicated in the former (but validation does not imply verification, to me).

ndushay:
The results of a F2F discussion with @jmartin-sul:

- stacks redirect param name will be `stacks_token`:
```
  redirect_to "#{@media.to_playlist_url}?stacks_token=#{encrypted_token}"
```
- the *action* name (and method name) in media controller will remain `verify_token`
- the param name for `verify_token` action will also be `stacks_token` to be consistent with the redirect param name
```
 (stacks_base_url)/media/id/filename.mp4/verify_token?stacks_token=encrypted_token&user_ip=192.168.1.100'
```

@jkeck we believe these are the only remaining open questions.  I will code to this spec and will then be ready for hopefully merging.

ndushay:
ok, @jmartin-sul , @jkeck   I think this refactoring might be ready for merging.   I couldn't get it down to 4 commits easily, but can rebase if someone thinks it's truly useful.

PR #48, opened 2016-05-10
Add an action to the media controller to check if a user can read media.

jkeck:


app/controllers/media_controller.rb
```diff
@@ -38,8 +38,25 @@ def verify_token
...
+    else
+      { status: :must_authenticate,
+        service: {
+          '@id' => iiif_auth_api_url,
+          'label' => 'Stanford-affiliated? Login to view'
+        }
+      }
+    end
+  end
+
```
ndushay: I kinda like having "allowed_params" be the first private method, and rescue_can_can second, but it's not like we adhere to a specific order.  What do you think?   I believe I made that true in the other controllers, FWIW.

spec/controllers/media_controller_spec.rb
```diff
@@ -195,4 +195,34 @@
...
+    end
+
+    context 'must authenticate' do
+      it 'returns json that indicates the user must authenticate' do
+        get :auth_check, id: id, file_name: file_name
+        body = JSON.parse(response.body)
+        expect(body['status']).to eq 'must_authenticate'
+      end
+
+      it 'indicates where the client can authenticate' do
```
ndushay: 😆 so basically this test is saying "it tells the user where to go" 😆

ndushay:
The code looks fine, with one minor line comment.   You may have explained this in standup when my hangout client was frozen, but:  why is it we need this special action for the media controller but we don't need more than cancan for the other controllers?

app/controllers/media_controller.rb
jkeck: I've moved it.

jkeck:
Something similar _kind of_ exists in the form of data in the IIIF response for images: https://github.com/sul-dlss/digital_stacks_rails/blob/master/app/controllers/iiif_controller.rb#L104-L107.  That is where I drew the inspiration for the `service`/`@id` json keys.

For files, there is no need to present a user with a link to authenticate in the way that an image or time-based media viewer would because the act of attempting to access the file (download) will present a webauth challenge if the content is restricted.

ndushay:
ah, ok.  Now I understand.

app/controllers/media_controller.rb
ndushay: 👍

ndushay:
👍  to merge from me.

Issue #51, opened 2016-05-13
Implement location-based restrictions in Digital Stacks


LynnMcRae:
Referencing https://github.com/sul-dlss/media/issues/55.

The Stacks must take the rights information from dor-rights-auth that include location-based restrictions and determine if a request for a resource is from a blessed IP address or host per mapping rules that use the information from #50.

ndushay:
Much progress has been made.  Still needing attention:  adjusting the `media_controller` method  `#json_for_media_auth`.  

There are basic pieces ready:
* dor-rights-auth  branch
* digital_stacks_rails branch

may need sul-embed piece, too???

jmartin-sul:
i believe there is still sul-embed UI refinement to be done (sul-dlss/sul-embed#584).  but the digital_stacks_rails and dor-rights-auth parts appear to be finished, so closing this for now.

PR #55, opened 2016-06-03
implement location based rights

ndushay:
connects to #51

depends on  https://github.com/sul-dlss/dor-rights-auth/pull/10

@jkeck  did the heavy lifting for the actual code; I did more with the tests.   We think it's possible this isn't correct yet, but this PR is already large (yet, most changes are required for this initial stab at location based rights).  

The intent is to get this merged and address whatever isn't working in additional PRs.  For example, we haven't tied this into sul-embed at all.  We haven't deployed this and poked around;  @jkeck may have run it locally, but I was mostly focused on figuring out the tests and getting them to work.  Etc.


jmartin-sul:
`MediaController` should have access to the user object, right?  maybe that's where it should get the IP address?  e.g.:
```
  def encrypted_token
    # we use IP from which request originated -- we want the end user IP, not
    #   a service on the user's behalf (load-balancer, etc.)
    StacksMediaToken.new(id, file_name, current_user.ip_address).to_encrypted_string
  end
```

that way, the logic for where the IP address comes from is centralized.

side-note:  discovered that you can't comment inline on lines that aren't part of the diff.  was mildly surprised.

spec/models/user_spec.rb
```diff
@@ -166,6 +166,46 @@
...
+            </rightsMetadata>
+          XML
+        end
+
+        it { is_expected.not_to be_able_to(:download, file) }
+        it { is_expected.not_to be_able_to(:download, image) }
+        it { is_expected.not_to be_able_to(:download, media) }
+      end
+    end
+
```
jmartin-sul: minor quibble:  seems like `rights_xml` is the same for both of these tests.  what about moving the `let(:rights_xml)` out into the wider `context 'location based'` instead of repeating it in each `anonymous user` context?  seems like that might make the test slightly clearer and slightly shorter?

jmartin-sul:
a few more notes:
* it seems like it'd be nice to check repsonse code and content type in the success cases of file_auth_request_spec (as is done in iiif_auth_request_spec).
* likewise, it'd be nice to check response code in success cases of media_auth_request_spec.
* i like how the first `"prompts for webauth when user not webauthed"` in media_auth_request_spec checks what happens after the redirect is done and the user logs in.  is it worth adding that to other `"prompts for webauth when user not webauthed"` tests?  or would that be overkill?
* i think i'm neutral on leaving the skipped tests.  i'm having a bit of a hard time telling whether they'd be usable as-is if we did implement `location AND stanford`.  if they're realistic for that situation, i'd leave them in.  if they feel like too much of a guess for how that machinery would work, i'd be inclined to take them out.

i could also use a quick walkthrough of the Ability model changes and the new ApprovedLocation model.  they seem pretty straightforward, just want to make sure i actually do understand the logic there.

jmartin-sul:
minor concerns, and i'd like to make sure i understand what's going on with the model changes, but i'm overall pretty much 👍 on this from what i can see, even as-is.  @jkeck mentioned this morning that he had some things he's inclined to change in retrospect, so might be good to meet and discuss those concerns, esp if they're different from the things i noticed.

ndushay:
Met with @jkeck and @jmartin-sul F2F, did a few tweak and explanations to address concerns.  Rebased.  I think this is ready for someone to push big green button?

Unless @jmartin-sul's comment has unfinished business:

```
 @jkeck mentioned this morning that he had some things he's inclined to change in retrospect, so might be good to meet and discuss those concerns, esp if they're different from the things i noticed.
```

Issue #49, opened 2016-05-12
ensure media (and other) filenames work with special chars


ndushay:
utf-8 chars, spaces, etc.  

Are they properly url encoded in redirects to wowza server?

ditto for encrypted stacks tokens

related to sul-dlss/dlss-wowza#9

See https://github.com/sul-dlss/media/wiki/Examples-of-files-deposited-by-ETD-users,-special-characters-and-media-files  for examples.

peetucket:
I believe stacks should work with space filenames - i ran into this issue several years ago with images for Revs and i think we fixed it (i.e. files with spaces in it should be served up by the stacks...though that was in the "old stacks code)

ndushay:
ensure that roundtrip of   stacks-redirect --> wowza --> stacks token verification -->  user   works for weird chars in filenames

jkeck:
FWIW, the `MediaController` uses the same routing mechanism as file/image, so if we support the URLs for downloading files in existing published PURLs, then they should already be supported on the digital stacks side.

We will want to ensure that the Wowza plugin  properly decodes/encodes the URLs that it receives/sends-out, but this should already be 👌 in stacks.

Any issues that come up would have already been issues with downloading the files (which I'm sure we would have heard about by now).

jmartin-sul:
@jkeck suggested during standup that if we wanted to test this, we could test it in the media controller routing spec

ndushay:
just for the record, @jkeck's comment of over 3 weeks ago was incorrect -- the routing for the media controller used diff constraints for filenames, so this really was necessary work.

Furthermore, if seemed no one knew the allowed charset even for non-media filenames, so various tickets have been spawned to pursue that.

PR #56, opened 2016-06-04
handle special chars in download and stream URLs, add tests for special char handling in URLs in general

jmartin-sul:
this builds on the work @ndushay started in 49-tests-for-special-chars-in-url. 

Note that it may not be perfectly what we want at this time, but it should be LESS BROKEN that our original filenames for media routes.

 it:
* alters a couple file name parameter constraints to allow for special char handling in media streaming and download routes.
* adds a note about a limitation on usable file names for the download routes.
* does some rubocop mandated style cleanup on the tests naomi already added for special character handling in general across media, file, and iiif routing specs.
* refines some skip messages on some of the tests naomi added.

connects to #49

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,13 +5,16 @@ Rails.application.routes.draw do
     get '/file/auth/:id/:file_name' => 'webauth#login_file', as: :auth_file
   end
 
-  constraints file_name: %r{[^/][\w-]+} do
+  # download file_name must NOT include format extension (e.g.  .mp4)
+  # so insist on . being URL encoded (see media routing specs)
+  constraints file_name: %r{[^/\.]+} do
     get '/media/:id/:file_name' => 'media#download', as: :media
     get '/media/auth/:id/:file_name' => 'webauth#login_media_download', as: :auth_media_download
   end
 
   # stream file_name must include format extension, eg .../oo000oo0000.mp4/stream
-  constraints file_name: %r{[^/][\w\.-]+} do
+  #  other dots do not need to be URL encoded (see media routing specs)
+  constraints file_name: %r{[^/]+\.\w+} do
     get '/media/:id/:file_name/stream' => 'media#stream', as: :media_stream
     get '/media/auth/:id/:file_name/stream' => 'webauth#login_media_stream', as: :auth_media_stream
     get '/media/:id/:file_name/verify_token' => 'media#verify_token'
--- a/spec/routing/file_routing_spec.rb
+++ b/spec/routing/file_routing_spec.rb
@@ -5,6 +5,49 @@ describe 'file routes' do
     expect(get: '/file/abc/def.pdf').to route_to('file#show', id: 'abc', file_name: 'def.pdf')
   end
 
+  context '#show (download): filename with' do
+    it 'chars not requiring URI escaping' do
+      filename = "(no_escape_needed):;=&$*.-_+!,~'.ext"
+      expect(get: "/file/oo000oo0000/#{filename}").to route_to(
+        'file#show', id: 'oo000oo0000', file_name: filename)
+    end
+
+    it 'some chars requiring URI escaping' do
+      filename = 'escape_needed {} @#^ %|"`.ext'
+      expect(get: "/file/oo000oo0000/#{URI.escape(filename)}").to route_to(
+        'file#show', id: 'oo000oo0000', file_name: filename)
+    end
+
+    it 'square brackets must be url escaped' do
+      filename = 'foo[brackets].bar.pdf'
+      escaped_filename = 'foo%5Bbrackets%5D.bar.pdf' # URI.escape doesn't do square brackets
+      expect(get: "/file/oo000oo0000/#{escaped_filename}").to route_to(
+        'file#show', id: 'oo000oo0000', file_name: filename)
+    end
+
+    it 'question mark must be url escaped' do
+      filename = 'foo?.pdf'
+      escaped_filename = 'foo%3F.pdf' # URI.escape doesn't do question mark
+      expect(get: "/file/oo000oo0000/#{escaped_filename}").to route_to(
+        'file#show', id: 'oo000oo0000', file_name: filename)
+    end
+
+    it 'ü must be url escaped' do
+      skip('problem writing test:  ü decodes to \xC3\xBC')
+      filename = "fü.pdf"
+      escaped_filename = URI.escape(filename) # becomes %C3%BC
+      expect(get: "/file/oo000oo0000/#{escaped_filename}").to route_to(
+        'file#show', id: 'oo000oo0000', file_name: filename)
+    end
+
+    it '%20 in actual name (from ck155rf0207)' do
+      skip('problem writing test:  %20 becomes space, over-aggressive param decoding?')
+      filename = 'ARSCJ%202008.foo.bar.pdf'
+      expect(get: "/file/oo000oo0000/#{URI.escape(filename)}").to route_to(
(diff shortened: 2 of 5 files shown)


config/routes.rb
```diff
@@ -5,13 +5,18 @@
     get '/file/auth/:id/:file_name' => 'webauth#login_file', as: :auth_file
   end
 
-  constraints file_name: %r{[^/][\w-]+} do
+  # known issue:  this will fail to match for file names with multiple dots, though
+  # the rails magic that populates `format` with the extension after the dot will work
+  # for files with one dot.  e.g. "/media/oo000oo0000/video.mp4" will work and have a
+  # `format` of "mp4", but "/media/oo000oo0000/adjective.video.mp4" will result in
+  # "No route matches".
+  constraints file_name: %r{[^/][\w\(\)\{\}\[\]\*\+\|\^ @:#%;=&$!,~"'_-]+} do
```
ndushay: if the test won't pass with square brackets, should we take them out of this?   

But really: why would this be different when the code is so similar to file#show ?   Maybe we can pair and poke around together?

spec/routing/media_routing_spec.rb
```diff
@@ -6,23 +6,111 @@
...
+    end
+
+    it 'some chars requiring URI escaping' do
+      filename = 'escape_needed {} @#^ %|"`'
+      expect(get: "/media/oo000oo0000/#{URI.escape(filename)}.mp4").to route_to(
+        'media#download', id: 'oo000oo0000', file_name: filename, format: 'mp4')
+    end
+
+    it 'square brackets must be url escaped' do
+      skip("problem writing test:  the route doesn't seem to match on brackets, even if they're escaped")
```
ndushay: why would this be different when the code is so similar to file#show ?   Maybe we can pair and poke around together?

also - it works for #stream, but not #download?  wacko.

jmartin-sul: i'm not sure, and also found this behavior surprising.  maybe the regex in the routes config isn't escaping the brackets correctly?  the regex for the show route accepts anything that's not a slash, whereas the regex for download and stream explicitly lists all of the characters it find acceptable.  maybe `\[` doesn't actually match a bracket.

spec/routing/file_routing_spec.rb
```diff
@@ -5,6 +5,49 @@
...
+    it 'ü must be url escaped' do
+      skip('problem writing test:  ü decodes to \xC3\xBC')
+      filename = "fü.pdf"
+      escaped_filename = URI.escape(filename) # becomes %C3%BC
+      expect(get: "/file/oo000oo0000/#{escaped_filename}").to route_to(
+        'file#show', id: 'oo000oo0000', file_name: filename)
+    end
+
+    it '%20 in actual name (from ck155rf0207)' do
+      skip('problem writing test:  %20 becomes space, over-aggressive param decoding?')
```
ndushay: does this happen in RL?  Is the problem with the *test*, or the problem with the *code* ?

ndushay:
@jmartin-sul:   I realized too late that I should have left your branch alone and worked in a diff branch - my apologies.  (Keep your local copy of this branch -- rename it before you play with my changes!)

However, success!
* the mystery of the square brackets test failing had to do with the period, not the square brackets -- test fixed.
* wrote tests to explicitly show behavior needed for dots
* greatly simplified regex for file_name for media actions
* got rid of dup tests (media routing tests for hyphen)

I think this is better now;  I hope you agree.

ndushay:
So for me, the only remaining question is about whether %20 ( and perhaps ü ) work in reality but we can't write tests, or whether they do NOT work in reality.   I *think* @LynnMcRae gave up on putting ü or other utf-8 in the filenames as the ingest process wouldn't take them or some such.  So I am mostly concerned with the actual object that has %20 in it.    (And all the other weird chars we will get but haven't anticipated)

ndushay:
Note for @LynnMcRae and @hannahfrost:  when we deploy this code somewhere and you can bang around on it, it would be good for documentation to indicate what are legal file names.

LynnMcRae:
The one example of %20 in a file name was an unpublished file, and latin1 foreign characters are not a requirement. I think we have gone well beyond the reasonable set -- a great thing where it takes nothing to make the allowed character set larger, but we can reasonably constrain file names that must participate in URL-based delivery. Should I take the chattering cat filename as the allowed set? @ndushay, are you asking about %20 as an implied encoded space, or is the % character itself a problem?

ndushay:
@LynnMcRae  I'm not sure about the % char outside of %20.  Our written tests would indicate % needs to be URL encoded when sending to digital stacks as part of a file name, but %20 is hopeless even with % encoded (%2520).

I think this PR is really about fixing my oversight with the original notion of media filenames.  True testing of the limits of our allowed filenames should go with sul-dlss/media#61.  

I don't think the chattering cat file name is going to be the allowed set -- I think manual testing is indicated for that one.  I suspect the file name trips up various parts of our stack that don't URL encode always.  This whole issue is sort of a side effect / distraction because I happened to be concerned that my filename constraints on media objects was too strict ... and it was, but is relaxed greatly with this PR.

jmartin-sul:
merged this after playing with the tests a bunch and hemming/hawing a bit.  but as @ndushay pointed out, this is an improvement over both the state of master and the state i'd gotten things to in what i pushed up, so seemed worth getting this in.

i have to say, it seems really weird to me that a caller is expected to encode all but the last dot (if any) in a file name being requested for download (though i noticed that when file name was encoded that way, as @ndushay did in her multiple dots test, multiple dots worked with the prior regex too).  e.g., in my experience, `file.something.ext` as a segment in a path should be one atomic chunk and all the dots should be encoded or not.  but also `file.something.ext` should elicit the same behavior as `file%2Esomething%2Eext` as `file%2Esomething.ext` (though only the last of these will actually work with the current routing, the other two will return `No route matches`).  though i can see how that'd be quite dependent on webserver and framework implementations, since some part of the stack has to recognize the difference btwn `%2E` and `.` so that the former can be decoded to the latter.

i'd be in favor of filing an issue to at least explore "fixing" that (assuming my expectation is reasonable, since it's just based on my own anecdotal experience and not an RFC or something).  it seems like stacks should either take the file name as one uniformly encoded chunk, or it should ask the caller to break down the file and the extension into separate parameters, but it shouldn't do the weird "encode the first n-1 dots thing" (which feels like an artifact of rails default magic colliding with a routing customization).  put another way, figuring out file extension seems like stacks' responsibility and not the caller's.  happy to let this lie if it seems like it's not a big deal to others (or waiting to do it in stacks if/when trying to do it in callers becomes onerous).

another couple stray encoding thoughts:
* this new code should allow the "Cat Noah" example to work, right?  i was going to say whatever's serving the link (sul-embed?) would have to encode the dots, which made me realize that what i said above was a bit off:  the caller doesn't always want to encode n-1 dots, they want to encode all the dots up to the one denoting an extension, if any.  but "Cat Noah" has no extension, so it seems it should be possible to serve it if all of its dots are encoded to `%2E`.
* i had a branch with some WIP code for manually parsing the extension from the filename and accepting dots in the file name.  that's the thing that i was working on between reading PRs yesterday, so i'll try to run it by @ndushay and @jkeck as soon as we have a chance today.  maybe after standup?

spec/routing/file_routing_spec.rb
jmartin-sul: definitely very interested to see whether this happens IRL, because i find it super-surprising.  and as i noted in my recent brain dump comment, the thing that you think would be a crude console approximation of what's happening does what you'd expect (URL encode the string, URL decode the string, get the string back).

Issue #54, opened 2016-06-03
jsonp from auth_check: stacks can allow stream but not download;  location as well as stanford restricted


ndushay:
The current implementation does not provide a way to separate "read" (i.e. stream) permissions from download permissions.   

Addressing this probably involves tweaking cancan ability class, and writing tests.  There may be a model in how the permissions work with "tile" for images.

peetucket:
i.e. rule=no download

ndushay:
check_auth jsonp  response only responses success or forbidden, but need nuances to tell sul-embed:

* location vs stanford only
* stream but not download

ndushay:
considering this closed.  if there are trailing tendrils of functionality needed, another ticket should be opened.

PR #61, opened 2016-06-21
media controller auth_check action

ndushay:
* can distinguish between location and stanford restrictions
* cancan distinguishes between read and download

connects to #54 

@jkeck  we have a number of questions for you.  Some TODOs are in the code already.   We may want to go through this list in person as a group.

* auth_check JSON returned
  * ok to consistently change `must_authenticate` to `stanford_restricted`? (also in sul-embed)
  * ok to return 2 status codes when object has both location and stanford restrictions?
  * ok that `label` is at top level  for location restricted, but also inside `service` for stanford restricted?
  * ok that `label` key is a string but others are symbols?
  * ok to separate the two lines in the label value so we can display them with diff font settings in sul-embed?

* ability.rb
  * is there a more idiomatic way to add the info here?
  * should we stop calling it `val` and instead use `stanford_restricted` or whatever is approp?
  * should we get some of those `val, rule` pairs outside of `can` blocks because we re-use them?
  * can you explain how the `access` method works at the bottom?  
    * is it essentially a private method used for assessing permissions for StacksImage?
    * `next true if blah` lines aren't obvious to me -- i tried to google to learn this syntax but failed in cancan context and in general ruby context

* are our tests complete and sufficient?

* not sure how sul-embed consumes "read but not download" files so don't know if we need to do more in auth_check for this  (we assumed no for now)  -- we only found auth_check request for stream action
  * how is the download shown in sul-embed (file list in download panel???) 
  * how does sul-embed determine which files to show


app/models/ability.rb
```diff
@@ -73,6 +86,21 @@ def initialize(user)
       can? :download, f
     end
 
+    can :read, StacksMediaStream do |f|
+      user_in_location, _rule = f.location_rights(user.location)
+      f.restricted_by_location? && user_in_location
+    end
```
jkeck: Is this functionally different than https://github.com/sul-dlss/digital_stacks_rails/blob/master/app/models/ability.rb#L66-L70 given https://github.com/sul-dlss/digital_stacks_rails/blob/master/app/models/ability.rb#L66-L70?

app/services/media_authentication_json.rb
```diff
@@ -0,0 +1,83 @@
+###
+# A class to model media authentication checks
+# as JSON for responding in a JSONP endpoint
```
ndushay: is this really returning JSON or a hash?  doesn't the controller turn it into JSON?

spec/services/media_authentication_json_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    before do
+      allow(media).to receive(:restricted_by_location?).and_return(true)
+      allow(media).to receive(:stanford_only_rights).and_return(true)
+    end
+
+    it 'returns JSON that indicates that the item is stanford restricted' do
+      expect(subject.as_json[:status]).to eq [:stanford_restricted, :location_restricted]
+    end
+
+    it 'retruns login service information' do
```
ndushay: typo:  retruns

spec/services/media_authentication_json_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    before do
+      allow(media).to receive(:stanford_only_rights).and_return(true)
+    end
+
+    context 'when the user is not Stanford authenticated' do
+      it 'returns JSON that indicates that the item is stanford restricted' do
+        expect(subject.as_json[:status]).to eq [:stanford_restricted]
+      end
+
+      it 'retruns login service information' do
```
ndushay: typo retruns

spec/controllers/media_controller_spec.rb
```diff
@@ -161,81 +161,4 @@
       end
     end
   end
-
-  describe '#hash_for_auth_check' do
-    context 'cancan authorizes' do
-      before { allow(controller).to receive(:can?).and_return(true) }
-      it 'returns hash with status of success' do
```
ndushay: i don't think we have a test for success anywhere if this is removed?  Maybe keep this and a test to show we call MediaAuthenticationJSON if cannot?

app/models/ability.rb
ndushay: Not sure.  Maybe take it out and see if tests pass?   I find it hard to think through the logic on its own.

spec/controllers/media_controller_spec.rb
jkeck: @ndushay https://github.com/sul-dlss/digital_stacks_rails/pull/61/files#diff-ca5485d067e4b47181ca4906ffede9f7L146 seemed sufficient.

app/models/ability.rb
jkeck: On closer inspection I believe I was miss-reading this.  I see that this is adding the read/stream exceptions from the download case.

PR #63, opened 2016-07-01
Set CORS headers for media.

tingulfsen:
This is half of the work for sul-dlss/sul-embed#581. The other half is to update the Dash player to the latest version.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -2,6 +2,8 @@
 # API for delivering streaming media via stacks
 class MediaController < ApplicationController
   before_action :load_media
+  before_action :set_origin_header, except: [:auth_check, :stream]
+  before_action :set_cors_headers, only: [:auth_check, :stream]
 
   rescue_from ActionController::MissingFile do
     render text: 'File not found', status: :not_found
@@ -46,6 +48,17 @@ class MediaController < ApplicationController
   end
 
   private
+
+  # We do not rely on the web server to set Access-Control-Allow-Origin for *any* /media request,
+  # so we set it manually ourselves.
+  def set_origin_header
+    response.headers['Access-Control-Allow-Origin'] = '*'
+  end
+
+  def set_cors_headers
+    response.headers['Access-Control-Allow-Origin'] = Settings.cors.allow_origin_url
+    response.headers['Access-Control-Allow-Credentials'] = 'true'
+  end
 
   def allowed_params
     params.permit(:action, :callback, :id, :file_name, :format, :stacks_token, :user_ip)
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -39,3 +39,6 @@ user:
   locations:
     spec:
       - an.ip.address
+
+cors:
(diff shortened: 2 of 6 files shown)


app/controllers/media_controller.rb
```diff
@@ -47,6 +49,17 @@ def auth_check
...
 
+  # We do not rely on the web server to set Access-Control-Allow-Origin for *any* /media request,
+  # so we set it manually ourselves.
+  def set_origin_header
+    response.headers['Access-Control-Allow-Origin'] = '*'
+  end
+
+  def set_cors_headers
+    response.headers['Access-Control-Allow-Origin'] = Settings.cors.allow_origin_url
+    response.headers['Access-Control-Allow-Credentials'] = Settings.cors.allow_credentials
```
jkeck: I can't think of the case where we would ever want to set this to anything other than `true`, so I wonder if this is a good use of a configuration.  It's not particularly harmful, but I would be concerned that somebody might take this to indicate there is a use for setting this to `false`.  Up to you.

jkeck:
@tingulfsen, out of curiosity, did you happen to try to write a request test for this (as opposed to a controller test)?  The reason I ask is because controller tests like this are going away in rails 5, so anything we can do to reduce the number of tests we'll eventually need to migrate in the future will be beneficial.

That shouldn't necessarily hold this PR up, just wanted to put that out there since rails 5 has now officially released.

config/settings/test.yml
```diff
@@ -1,6 +1,10 @@
 app_users:
   test-user: 'password'
 
+cors:
+  allow_origin_url: 'https://embed-dev.stanford.edu'
```
jkeck: Any setting that we add should have a default setting in `settings.yml`, and then environmental exceptions should be placed in the environment specific `.yml` file.

I think we could simply move this to `settings.yml` and let that point to `embed.stanford.edu`.  Then if we need to configure it differently we can do so a `.local.yml` file (or on the server in the appropriate environment file.

This would mirror how we configure some of our other public services in this repo.

jkeck:
So, in testing this in the real world, I see that when a site like PURL embeds SUL Embed (natively and not through an iframe) the Origin is actually PURL and not Embed.  This means that the request still cannot send the credentials.

@tingulfsen, I think your solution about checking the incoming `Origin` and checking it against a whitelist before setting the `Access-Control-Allow-Origin` may be an interesting one, but it still presents the problem that it will not allow outside consumers to embed the HTML that we provide directly (since the origin might be somebody's wordpress blog).

Let's chat about potential solutions when we're all together next week.

config/settings.yml
```diff
@@ -39,3 +39,6 @@ user:
   locations:
     spec:
       - an.ip.address
+
+cors:
+  allow_origin_url: 'https://embed-dev.stanford.edu'
```
jkeck: Can we make this be just `embed.stanford.edu` as the default behavior (and we'll add exceptions to the appropriate settings files when necessary)?

It seems like the other things services (like PURL and ImageServer) all have their production URLs in this file, so it falls in line with the pattern that's already established.

PR #64, opened 2016-07-12
It is possible (when going through a proxy) to have multiple IPs…

jkeck:
…set, when ends up getting a comma separated list. The last one should be the user's real IP.

cc: @jmartin-sul, this is what we were talking about earlier.  Let me know if this looks reasonable.

### DIFF
--- a/app/models/stacks_media_token.rb
+++ b/app/models/stacks_media_token.rb
@@ -55,7 +55,7 @@ class StacksMediaToken
   def initialize(id, file_name, user_ip)
     @id = id
     @file_name = file_name
-    @user_ip = user_ip
+    @user_ip = last_ip_address(user_ip)
     @timestamp = Time.zone.now
     validate
   end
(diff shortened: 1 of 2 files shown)


app/models/stacks_media_token.rb
```diff
@@ -55,7 +55,7 @@ def self.max_token_age
   def initialize(id, file_name, user_ip)
     @id = id
     @file_name = file_name
-    @user_ip = user_ip
+    @user_ip = last_ip_address(user_ip)
```
jmartin-sul: if we do go with this approach, i'd prefer to rename `user_ip` to something more accurate like `user_ip_list` (or the verbose but even more accurate `user_ip_list_string`).

but my gut reaction is that this splitting logic should live in consumers of `StacksMediaToken`, and not `StacksMediaToken` itself, since the token class should (it seems to me) concern itself with the user's actual IP, and not the list of IPs that the caller gets from the request headers (that feels very controller-ish, since it's pretty specific to request handling).  that said, i wouldn't lose too much sleep over this living in `StacksMediaToken`, under the reasoning that all consumers of `StacksMediaToken` will want this functionality in practice.

jmartin-sul:
i'd be 👌 with this for expediency's sake, but left a suggestion inline since that didn't feel to me like an ideal place for that logic to live.

jkeck:
Closing this since the work will actually be completed in sul-dlss/dlss-wowza#27

PR #73, opened 2016-07-28
Remove Webmock, VCR, and all cassettes.

### LINKED ISSUE #40: remove VCR and webmock to lower maint costs
possibly only used for calls to purl - could use mocks instead.

jkeck:
Closes #40


jkeck:
Hmm, I don't think the code paths that coveralls is complaining about were *actually* tested before, it was just that the code paths happened to be executed.

That being said, I want to take a look at writing some tests for the areas that were not tested before (and we didn't know until this change).

cbeer:
Yeah, they were incidentally tested before, so requests for e.g. the Djatoka metadata would go through the stack and hit the VCR cassette, while now they're getting stubbed higher up.

ndushay:
:+1:  :+1:  :+1:  :+1:  :+1:  :+1:  :+1:  :+1:

jmartin-sul:
i second @ndushay's enthusiastic 👍

PR #75, opened 2016-08-01
Add a few tests now that we stub PURL XML a bit higher up the stack

jkeck:
I was working on these tests to be included with #73, but it was merged before I had a chance to finish them up.

### DIFF
--- a/spec/models/djatoka_metadata_spec.rb
+++ b/spec/models/djatoka_metadata_spec.rb
@@ -1,0 +1,63 @@
+require 'rails_helper'
+
+##
+# A stub test class to model
+# a Djatoka::Metadata class
+class DjatokaMetadataTestClass
+  def height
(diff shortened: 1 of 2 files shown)


spec/models/djatoka_metadata_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  describe '#max_height' do
+    it 'is the height from the metadata as an integer' do
+      expect(subject.max_height).to be 200
+    end
+  end
+
+  describe '#max_width' do
+    it 'is the height from the metadata as an integer' do
```
jmartin-sul: test description should probably say "width" instead of "height"?

jkeck: 👍 Damn copy-pasta.

PR #81, opened 2016-08-16
Add feature flags for location based auth and streaming media.

### LINKED ISSUE #80: Move media related work behind simple feature flag
We may be able to get a version of stacks that does not have media enabled (so we can release to production at-will) by adding a simple `RailsConfig` setting to enable/disable media behavior.

I think if we put the following behind a feature flag we can release to production w/ minimal regression testing required:

- [ ] [Routing to the media controller](https://github.com/sul-dlss/digital_stacks_rails/blob/master/config/routes.rb#L10-L22)
- [ ] [Location based auth 1](https://github.com/sul-dlss/digital_stacks_rails/blob/master/app/models/ability.rb#L44-L51)
- [ ] [Location based auth 2](https://github.com/sul-dlss/digital_stacks_rails/blob/master/app/models/ability.rb#L79-L83)
- [ ] [Location based auth 3](https://github.com/sul-dlss/digital_stacks_rails/blob/master/app/models/ability.rb#L89-L92)

Other things that I considered but I believe will actually be 👌 to :shipit: are the `AnonymousLocatableUser` change (as we have already validated that it works with the existing image viewer code) and other references to `StacksMediaStream` in the `Ability` class (given that `StacksMediaStream` is only initialized in the `MediaController` which will not be routable).

jkeck:
Closes #80 

I believe with this PR, master has only refactors and dependency updates in routable code.


cbeer:
Not sure what to make of the build failure (bad test, maybe?), but we can drop the Ruby 2.2 build if that helps.

ndushay:
LGTM.  👍

ndushay:
I think the failing build was just "one of those things".

PR #89, opened 2016-08-17
Remove code to generate media URLs and redirect requests to them.

### LINKED ISSUE #88: Remove StacksMediaStream code
The logic for generating URLs to the streaming server now resides in embed.  We should remove this code from stacks now that we're confident in our approach.  Some of the things that can be removed are the following:
- [x] Remove the model code responsible for generating the URL to the streaming media server
- [x] Remove the controller action that handles redirecting to the streaming media server
- [x] Remove `is_it_working` check for streaming server

jkeck:
Closes #88


jkeck:
I believe the code coverage reduction is due to dropping in relevant lines of code tested given that this PR remove > 400 lines.

ndushay:
LGTM at a quick glance;  I want to review it more carefully to see if I can think of anything else to come out (or someone else can do this).

ndushay:
Some additional spots to be removed?

- [ ] in media_controller.rescue_can_can  line `if allowed_params['action'] == 'stream'` ...
- [ ] should `spec/requests/media_auth_request_spec`  be renamed `location_file_auth_request` spec or combined with `file_auth_request_spec` or some such?

jkeck:
@ndushay good catch on those.  I removed the reference to the stream action in the `rescue_can_can` method.  WRT the test, I've added a comment to #92 about doing that as part of that work since whoever does that might have a better idea of where that logically fits.

PR #90, opened 2016-08-18
Replace Squash with Honeybadger for exception notification.

### LINKED ISSUE #79: add honeybadger to stacks
We want non-failing notifications to cause reporting.

- want to notify for unparseable durations

For a ruby example, see sul-dlss/exhibits#291

once done, remove squash

jkeck:
Closes #79 

See sul-dlss/shared_configs#104 & sul-dlss/shared_configs#105 for honeybadger configs.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -24,7 +24,7 @@ require 'capistrano/deploy'
 require 'capistrano/rails'
 require 'capistrano/passenger'
 require 'dlss/capistrano'
-require 'squash/rails/capistrano3'
+require 'capistrano/honeybadger'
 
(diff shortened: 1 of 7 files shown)


ndushay:
LGTM;  not sure if we can do this yet (if honeybadger account allows enough applications)

jkeck:
Ah, sorry, I guess I didn't explain that well enough in standup.  The account issue was due to the fact that the apps were under me and not our team.  Transferring them over to our team solved this issue.

ndushay:
I'm good with this PR as is, but do we want honeybadger notification for anything else besides the one exception?  Should that be a separate ticket?  etc.

PR #94, opened 2016-08-22
Remove MediaController#download in favor of FileController#show

### LINKED ISSUE #92: Investigate if we need a download action in the media controller
I don't believe this is routed anywhere currently. File download happens through the file controller at `/file/{druid}...`.

Let's see if we can KWF.

jkeck:
Closes #92


app/models/ability.rb
```diff
@@ -46,22 +46,21 @@ def initialize(user)
...
-
         # location OR stanford, not location AND stanford
         f.restricted_by_location? && (!user_in_location || !rule.blank?)
       end
     end
 
     # media download rights exclude when rule is no-download
     # stanford restricted, read, but no download
-    cannot :download, StacksMediaStream do |f|
+    cannot :download, [StacksFile, StacksImage, StacksMediaStream] do |f|
```
ndushay: should comment in line 55 be changed to not have the word "media" any more?

app/models/ability.rb
```diff
@@ -46,22 +46,21 @@ def initialize(user)
...
       stanford_only_rights, rule = f.stanford_only_rights
       stanford_only_rights && rule == 'no-download'
     end
 
     # media download rights exclude when rule is no-download
```
ndushay: ditto for this comment ...

config/routes.rb
```diff
@@ -6,14 +6,7 @@
...
-    # download file_name must NOT include format extension (e.g.  .mp4)
-    # so insist on . being URL encoded (see media routing specs)
-    constraints file_name: %r{[^/\.]+} do
-      get '/media/:id/:file_name' => 'media#download', as: :media
-      get '/media/auth/:id/:file_name' => 'webauth#login_media_download', as: :auth_media_download
-    end
-
-    # stream file_name must include format extension, eg .../oo000oo0000.mp4/stream
+    # stream file_name must include format extension, eg .../oo000oo0000.mp4/verify_token
     #  other dots do not need to be URL encoded (see media routing specs)
```
ndushay: This comment made me think.  The specs that were checking for special chars have been removed.  (see https://github.com/sul-dlss/digital_stacks_rails/commit/287304de5adb0f5190fce79f2b1c4d97d9b711e7)   I wonder if we know what we need for stream name encoding for verify_token, and if we should have routing tests to illustrate it.  Or if we should remove this comment.   I also put a reference in sul-dlss/media#111

spec/requests/media_auth_request_spec.rb
```diff
@@ -2,17 +2,11 @@
 
 RSpec.describe "Authentication for Media requests", type: :request do
 
-  let(:allowed_loc) { 'ip.address1' }
   let(:user_no_loc_no_webauth) { User.new }
-  let(:user_no_loc_webauth) { User.new(webauth_user: true) }
```
ndushay: The changes here confuse me a bit.  Do we no longer care about:
* various user scenarios when an object is both location and stanford restricted. 
* the user scenario when user is webauthed but not stanford.

Or, is that not testable here any more, but is tested somewhere else?   Because we certainly will still have the possibility of all these types of users interacting with stanford-only, with location-only and with both for streaming ... and I assume that is still handled by stacks?

jkeck: I believe that this behavior is tested (multiple times) between the ability tests (via user spec), the controller tests, the request tests, and the unit tests for the `MediaAuthenticationJSON`. As opposed to removing this test entirely, I thought it was useful for indicating that the JSON was being served up correctly under the broad use cases (and the individual use cases are tested in the other places).

If we retained all of the scenarios here, it felt like we were unnecessarily duplicating assertions already being made.

config/routes.rb
jkeck: Ah, I was thinking that there were additional tests in that file for the other routes.  I'll add routes back in that check one of the other paths so that can keep those checks in somewhere.

ndushay: you mean you'll add similar tests for special chars for the verify_token route, that used to exist for the other routes?

spec/requests/media_auth_request_spec.rb
ndushay: I thought the request specs were sort of the "full stack" spec and it was appropriate to have them here when we had media stream and download actions.   Now that we don't have the media stream action ... where are we asserting these interactions for permissions for streaming requests?    I may be just missing where these interactions are tested in the user spec ... but I have looked there and elsewhere and I'm not seeing the full interactions.

Basically, I may be confused as to where we really are implementing location-based permissions for streams:  sul-embed calls stacks for ?? file read??  and that is where this is now handled for streams?    If so, the file request specs don't handle the case for user webauthed but not stanford ... or doesn't that matter?  Or am I just confused?

I have looked in all the spots you indicated above ... and still feel a bit concerned that we are losing a few test cases along the way.

jkeck: The CanCan ability for non-stanford WebAuth users is tested here: https://github.com/sul-dlss/digital_stacks_rails/blob/92-remove-media-download/spec/models/user_spec.rb#L96-L151 (bonus stream vs. download tests all over the place in that file).

This action is an implementation of `can? :stream, StacksMediaStream`.  I feel like the various user/rights scenarios are tested thoroughly between the spec above and the [`FileAuthRequestSpec`](https://github.com/sul-dlss/digital_stacks_rails/blob/92-remove-media-download/spec/requests/file_auth_request_spec.rb).

This test is doing some high-level tests to execute the code paths of 1) if a user can stream and 2) if a user can't stream.  The details of what makes a user be able to stream or not is tested elsewhere.  Also, within the `user can't stream` conditional branch, I feel that the `MediaAuthenticationJSON` thoroughly tests the scenarios of different user and file objects.

I would pose this question:  If we add a new file type with its own controller, and it implemented an action that checked that a user could read a file (which is already thoroughly tested), would we then need to setup every permutation of user and rights to test the results of that action?  That would seem unnecessary to me (and potentially be an undue increase in maintainability burden).

That being said, I would be more than happy to sit down with you tomorrow, and maybe we can hunt down some scenarios that I've missed in this refactor and we can make sure that they are tested appropriately.

ndushay: Thanks for that explanation.  I think I'm more comfortable now that I understand better what is tested where.  There is still a lot of magic in rails and can can that I don't always follow, and I think some of the original testing being removed here
- was about a then new action on a then new controller
- was about ensuring all the rails and cancan magic worked the way it's supposed to.

You are more conversant in both Rails and cancancancancancan so it probably made more sense to you from the get go.

I am now 👍 on this point being addressed.

ndushay:
👍 on merge.

PR #95, opened 2016-09-07
Hardcode response format in auth_check, to please IE/Edge on Windows 10.

tingulfsen:
For Windows 10, Internet Explorer 11 and Edge seem to request the wrong content type when the jQuery Ajax request to the auth_check endpoint is issued (this happens when Sul-Embed tries to play a video or audio file). Since we actually only ever return JSON from auth_check, this PR simply hardcodes the response format without testing the actual format requested. This is an easy and seemingly foolproof way to please all browsers. This is the stacks-side of the fix for sul-dlss/media#122.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -22,9 +22,9 @@ class MediaController < ApplicationController
 
   # jsonp response
   def auth_check
-    respond_to do |format|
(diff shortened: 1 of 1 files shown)


jkeck:
👍 LGTM, but we paired on this change.

Issue #96, opened 2016-09-13
stacks: okcomputer to replace is_it_working


ndushay:
Exemplar:  purl-fetcher

possibly:
- `/status` endpoint for load-balancer needs and nagios errors
- `/status/[some-dependency]` for nagios warnings;  nagios routes them to approp parties
- `/status/all-dependencies` for full dependency documentation (?)

after this
- [x] PR merged to master (#97, #99, #101)
  - [x] deploy
- [x] update Load Balancer (SAR-3254)
- [ ] update nagios 
  - [x] basic checks (~~sul-dlss/puppet#1014~~, ~~sul-dlss/puppet#1015~~, ~~sul-dlss/puppet#1016~~, ~~sul-dlss/puppet#1017~~) 
  - [x] stacks-lb (sul-dlss/puppet/pull/1024)
  - [ ] any individual checks need separate nagios handling? (punt to #86)
- [x] remove is_it_working (#105)
  - [ ] deploy
    - [x] dev
    - [x] stage
    - [ ] uat
    - [x] prod
- [ ] update DevOpsDocs (sul-dlss/DevOpsDocs/pull/62)

ndushay:
I'm closing this - the only remaining unchecked boxes either have separate issues or PRs.

PR #97, opened 2016-09-20
add app monitoring with okcomputer at /status

ndushay:
connects to #96

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -1,0 +1,52 @@
+require 'okcomputer'
+
+# /status for 'upness' (Rails app is responding), e.g. for load balancer
+# /status/all to show all dependencies
+# /status/<name-of-check> for a specific check (e.g. for nagios warning)
+OkComputer.mount_at = 'status'
+OkComputer.check_in_parallel = true
+OkComputer::Registry.deregister "database" # don't check (unused) ActiveRecord database conn
+
+# TODO: remove DirectoryCheck here after there is comparable functionality in okcomputer
+class DirectoryCheck < OkComputer::Check
+  attr_reader :path, :options
+  def initialize(path, options = {})
+    @path = Pathname(path.to_s)
+    @options = options
+  end
+
+  def check
+    mark_message "Directory check for #{path}: #{options.inspect}"
+    mark_failure if options[:read] && !path.readable?
+    mark_failure if options[:write] && !path.writable?
+  end
+end
+
+# REQUIRED checks, required to pass for /status/all
+#  individual checks also avail at /status/<name-of-check>
+OkComputer::Registry.register 'ruby_version', OkComputer::RubyVersionCheck.new
+# TODO: add app version check when okcomputer works with cap 3 (see http://github.com/sportngin/okcomputer#112)
+
+# Check the memcache servers used by Rails.cache
+if Rails.cache.respond_to? :stats
+  OkComputer::Registry.register 'rails_cache', OkComputer::CacheCheck.new
+else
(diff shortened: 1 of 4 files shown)


config/initializers/okcomputer.rb
```diff
@@ -0,0 +1,51 @@
...
+
+# Check the memcache servers used by Rails.cache
+if Rails.cache.respond_to? :stats
+  OkComputer::Registry.register 'rails_cache', OkComputer::CacheCheck.new
+else
+  OkComputer::Registry.register 'rails_cache', OkComputer::GenericCacheCheck.new
+end
+
+OkComputer::Registry.register 'stacks_mounted_dir',
+  DirectoryCheck.new(File.join(Settings.stacks.storage_root, 'bb'), read: true, write: true)
```
tingulfsen: Is there any reason you're not just using stacks.storage_root here on its own?

spec/features/status_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  it '/status checks if Rails app is running' do
+    visit '/status'
+    expect(page.status_code).to eq 200
+    expect(page).to have_text('Application is running')
+  end
+  it '/status/all checks if required dependencies are ok and also shows non-crucial dependencies' do
+    visit '/status/all'
+    expect(page).to have_text('purl_url')
+    expect(page).to have_text('imageserver_url')
+  end
```
tingulfsen: Would you expect a directory check message here also?

config/initializers/okcomputer.rb
jmartin-sul: maybe just making sure the first part of the druid tree is there for a known druid?  either way, an explanatory comment might be nice.

jmartin-sul:
definitely 👍 once tommy's questions get answered, but👌 overall even as-is.

spec/features/status_spec.rb
ndushay: yeah, but i'm really checking for a single required check and for an optional check.  No need to prove that okcomputer works as advertised.

config/initializers/okcomputer.rb
ndushay: it's referred to as `Settings.stacks.storage_root` in the code.   

```
[ndushay@m3dl-sm-04-mbph digital_stacks_rails (honeybadger-2.6.1)]$ ack storage_root
app/models/stacks_file.rb
36:      File.join(Settings.stacks.storage_root, match[1], match[2], match[3], match[4], file_name) if match && file_name

config/initializers/is_it_working.rb
2:  h.check :directory, path: File.join(Settings.stacks.storage_root, 'bb')

config/settings.yml
2:  storage_root: /stacks

spec/models/stacks_file_spec.rb
8:      expect(subject.path).to eq "#{Settings.stacks.storage_root}/ab/012/cd/3456/def.pdf"

spec/models/stacks_image_spec.rb
88:      expect(subject.path).to eq "#{Settings.stacks.storage_root}/ab/012/cd/3456/def.jp2"
```

jmartin-sul: ah, sorry, i'm not sure about tommy, but my question was about the use of `'bb'` in the check (not the use of `Settings` to refer to the setting).  as opposed to just having something like:
```ruby
DirectoryCheck.new(Settings.stacks.storage_root, read: true, write: true)
```

ndushay: that's what was in `is_it_working` check.   I assumed it was to ensure that the directory mount really was to the appropriate druid tree level.

spec/features/status_spec.rb
ndushay: i could put a comment to that effect at the end of each of those lines if that would be more clear

ndushay:
duly commented per suggestions.

PR #104, opened 2016-09-22
stacks_file: remove valid? method as it made no sense

### LINKED ISSUE #32: remove(?) stacks_file model's valid method
current code in StacksFile:
```ruby
def valid?
    # FIXME: image_exist? is defined in app/models/concerns/djatoka_adapter.rb,
    #  which is included in StacksImage, but is probably undefined here
    image_exist?
end
```

it's scary to mess with a method called "valid" in a model -- lots of rails magic may be at work. Probably the fix is to change it to file_exist?  

Coverage shows that line is never executed by our tests ...  could we remove the entire method?  etc.

ndushay:
I suspect this method was vestigial from some copy-paste that went bad, or something.

resolves #32

### DIFF
--- a/app/models/stacks_file.rb
+++ b/app/models/stacks_file.rb
@@ -15,12 +15,6 @@ class StacksFile
     path && File.exist?(path)
   end
 
-  def valid?
-    # FIXME: image_exist? is defined in app/models/concerns/djatoka_adapter.rb,
-    #  which is included in StacksImage, but is probably undefined here
-    image_exist?
-  end
-
   def mtime
     @mtime ||= File.mtime(path) if file_exist?
   end


app/models/stacks_file.rb
```diff
@@ -15,12 +15,6 @@ def file_exist?
     path && File.exist?(path)
   end
 
-  def valid?
-    # FIXME: image_exist? is defined in app/models/concerns/djatoka_adapter.rb,
-    #  which is included in StacksImage, but is probably undefined here
-    image_exist?
-  end
```
jkeck: I'm 👌 w/ removing this method altogether.

It feels a bit odd to have a subclass define a method that the parent class does not have base behavior for (but that seems to be a pattern in the `StacksImage` class, so :shipit:)

ndushay: everything about this method feels odd.  There is a separate file_exist?  method ... and I can imagine how rails magic might be happier with a "valid?" method that calls file_exist? ... but we don't seem to use valid?  and it certainly makes no sense as it was.

jkeck: Oh, totally agree.  Just using the opportunity to have a discussion about code. 😎

Some things to consider in a base class like this would be to set a default `true` or `false`.  Depending on the use of your `valid?` method in the app this could have potentially dangerous results.

Another thing I've seen which I like is to `raise NotImplementedError`.  That way if I subclass this, and the `valid?` method gets called anywhere in my routable code path (say I copy the IIIFController for some reason) then a nice big error is raised and you'll see it or it'll be caught by the exception handler.

ndushay: yeah - those are good suggestions for when we are trying to do something useful with valid?

PR #107, opened 2016-09-27
Update to Rails 5

cbeer:


jkeck:
Thanks for doing this @cbeer.  Since we're currently in a code-freeze due to active UAT, we'll hold off on getting this merged in until after we've deployed the media stacks (just in case the upgrade causes any unexpected regressions).

We'll discuss how to get this deployed out to production after we lift our code-freeze.

cbeer:
Thanks; there's obviously no hurry on this (other than general good practice of keeping dependencies up to date), just another easy app to tackle.

app/controllers/webauth_controller.rb
```diff
@@ -18,10 +18,10 @@ def login
   # TODO: can't think of a reasonable way to do strong params here
 
   def login_file
-    redirect_to file_path(params.symbolize_keys)
+    redirect_to file_path(params.to_unsafe_h.symbolize_keys)
   end
 
   def login_iiif
-    redirect_to iiif_path(params.symbolize_keys)
+    redirect_to iiif_path(params.to_unsafe_h.symbolize_keys)
```
jkeck: Is this because there are no strong params set in this controller?  If not, is there a reason we would use `to_unsave_h` over `to_h`?  (I'm a bit naive about this change in rails 5).

jkeck:
👍 LGTM.  One minor comment (that is most likely due to my ignorance).

If any other folks want to have a look before we :shipit:, please feel free.

Other than that, perhaps we can try and get this deployed out soon (as this is a relatively straight forward rails app and this update is pretty low-risk).

app/controllers/webauth_controller.rb
cbeer: Exactly; future work should include using strong params instead. In the meantime, we'll at least be aware of the danger (if any; this usage seems fairly innocuous, I think)

PR #109, opened 2016-09-29
Check status/default.json on PURL, instead of trying to load the home page

cbeer:


ndushay:
D'oh - what was I thinking ...

PR #115, opened 2017-01-06
Replace unmaintained Hurley gem with a combination of Faraday and HTTP

cbeer:
Deployed to -stage and seems to work fine.


app/models/concerns/djatoka_adapter.rb
```diff
@@ -2,14 +2,8 @@
 # Djatoka-backed implementation of StacksImage delivery
 module DjatokaAdapter
   def response
```
jcoyne: This is never called with a block?

app/models/concerns/djatoka_adapter.rb
```diff
@@ -2,14 +2,8 @@
...
-    return to_enum(:response) unless block_given?
-
     benchmark "Fetch #{url}" do
-      client.get(url) do |req|
-        req.on_body do |_, chunk|
-          yield chunk
-        end
-      end
+      HTTP.get(url).body
```
jkeck: Why `HTTP` here instead of `Faraday`?

cbeer: `HTTP::Reponse#body` streams the response back as an IO-like object. Faraday doesn't (yet) support content streaming.

app/models/concerns/djatoka_adapter.rb
jcoyne: Add documentation that this method is returning an IO.

app/models/concerns/djatoka_adapter.rb
jkeck: Thanks for the clarification.

app/models/concerns/djatoka_adapter.rb
cbeer: Done.

PR #132, opened 2017-06-14
Refactor authorization for understandability 

cbeer:


spec/models/user_spec.rb
```diff
@@ -315,12 +315,12 @@
         it { is_expected.not_to be_able_to(:read, file) }
         it { is_expected.to be_able_to(:read, tile) }
         it { is_expected.not_to be_able_to(:read, image) }
-        it { is_expected.not_to be_able_to(:stream, media) }
+        it { is_expected.to be_able_to(:stream, media) }
```
ndushay: why is streaming permission changing?

app/models/ability.rb
```diff
@@ -19,119 +19,60 @@ def initialize(user)
...
     #
     # The third argument is an optional hash of conditions to further filter the
-    # objects.
-    # For example, here the user can only update published articles.
-    #
+    # objects. For example, here the user can only update published articles.
     #   can :update, Article, :published => true
     #
+    # The block argument takes as a parameter an instance of the object for which
+    # permission's being checked. If the block returns true, the user is granted that
```
ndushay: i think it would be better to say `permission is` here just to be even clearer (as the added commas above make things clearer)

spec/models/user_spec.rb
cbeer: We think this was copy-pasta with a rights scenario that wasn't accounted for, and reviewed the change with @LynnMcRae.. In short, we believe `no-download` resources should be streamable (but not downloadable).

LynnMcRae: +1

ndushay: Gotcha.  Good to know.

app/models/ability.rb
ndushay: So if you take this suggestion, I would be happy to hit the big green button.

cbeer: @jmartin-sul I think this was from your commit; would you care to fix it?

jmartin-sul: yup, will do...

ndushay:
These changes have been confirmed as ok via manual testing.

app/models/ability.rb
jmartin-sul: fixed

PR #133, opened 2017-06-14
Advertise support for restricted locations; fixes #125

### LINKED ISSUE #125: Add authentication service description for location-based restricted access
(Presumably based on http://iiif.io/api/auth/1.0/#external-interaction-pattern; related to sul-dlss/purl#134)

cbeer:
(Includes #132)

### DIFF
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -60,8 +60,7 @@ class IiifController < ApplicationController
   #   a)  access not allowed (send to super)  OR
   #   b)  need user to login to determine if access allowed
   def rescue_can_can(exception)
-    stanford_restricted, _rule = current_image.stanford_only_rights
-    if stanford_restricted && !current_user.webauth_user?
+    if current_image.stanford_restricted? && !current_user.webauth_user?
       redirect_to auth_iiif_url(allowed_params.to_h.symbolize_keys.tap { |x| x[:identifier] = escaped_identifier })
     else
       super
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -114,23 +114,41 @@ class IiifController < ApplicationController
 
     info['sizes'] = [{ width: 400, height: 400 }] unless current_image.maybe_downloadable?
 
+    services = []
     if anonymous_ability.cannot? :download, current_image
-      info['service'] = {
-        '@id' => iiif_auth_api_url,
-        'profile' => 'http://iiif.io/api/auth/1/login',
-        'label' => 'Stanford-affiliated? Login to view',
-        'service' => [
-          {
-            '@id' => iiif_token_api_url,
-            'profile' => 'http://iiif.io/api/auth/1/token'
-          },
-          {
-            '@id' => logout_url,
-            'profile' => 'http://iiif.io/api/auth/1/logout',
-            'label' => 'Logout'
-          }
-        ]
-      }
+      if current_image.stanford_restricted?
+        services << {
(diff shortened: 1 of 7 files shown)


spec/controllers/iiif_controller_spec.rb
```diff
@@ -165,6 +186,47 @@ def stanford_restricted?
...
+        before do
+          allow(stub_metadata_object).to receive(:restricted_by_location?).and_return(true)
+          allow(stub_metadata_object).to receive(:restricted_locations).and_return([:spec])
+        end
+
+        context 'and the user is not in that location' do
+          it 'advertises an authentication service' do
+            expect(image_info['service']).to be_present
+            expect(image_info['service']['profile']).to eq 'http://iiif.io/api/auth/1/external'
+          end
```
jmartin-sul: might be nice to also check `'label'` and/or `'failureHeader'` for the expected text?  though i don't think that's super-important, and wouldn't object if you wanted to leave it out.   would also be nice to test `'failureDescription'` to see that it includes the expected location, though the method that gets the location list at least seems to be tested in `stacks_rights_spec.rb`.

jmartin-sul:
also, when i described this work to @jkeck yesterday, he expressed some concerns about caching of generated manifests (e.g. user accesses a manifest from a disallowed location, result gets cached, user accesses from allowed location, gets stale copy with location restriction message; or vice versa).  i didn't know enough about the mechanics of the caching in question to have any real insight one way or the other.

cbeer:
We only rely browser caching for `info.json` responses (for not very long), and bust those caches based on the current user. I believe for location-based restrictions, those are for physical kiosks which won't get shared across sessions anyway.

jkeck:
Ah, okay, I was def. thinking more of the presentation API manifest (which actually comes from PURL).

So, just for my personal edification, the same image API `info.json` reference is returned by the presentation manifest, but the returned image API `info.json` content will be different based on if the user is in the blessed location or not?

cbeer:
Yes, exactly.

jmartin-sul:
pushed a commit with the suggestions i'd had

jmartin-sul:
rebased on master and pushed updated branch (now that #132 has been merged)

jmartin-sul:
also, squashed my commit into @cbeer's last commit

app/controllers/iiif_controller.rb
```diff
@@ -115,23 +114,41 @@ def image_info
...
+          ]
+        }
+      end
+
+      if current_image.restricted_by_location? && !current_image.location_rights(current_user.location).first
+        services << {
+          'profile' => 'http://iiif.io/api/auth/1/external',
+          'label' => 'External Authentication Required',
+          'failureHeader' => 'Restricted Material',
+          'failureDescription' => "This material is only viewable on-site at #{current_image.restricted_locations}."
```
ndushay: how was the descriptive text arrived at?  Should it match what's in sul-embed (or wherever the text displayed in the viewer comes from)?

https://github.com/sul-dlss/sul-embed/blob/fb5fdb310aac598c11d851a1f37e5b6b3f0328c3/lib/embed/media_tag.rb#L112-L124

cbeer: Ah, good catch. We just made that text up.

jmartin-sul: will do...

jmartin-sul: i noticed that the embed text specifically mentions "playing" media, so i got advice from @jvine on more generic text for the message.

fixed and pushed.

PR #141, opened 2017-06-26
Support the IIIF 2.1 max size parameter

### LINKED ISSUE #139: Support IIIF 2.1 'max' image size parameter
http://iiif.io/api/image/2.1/#size

cbeer:
Fixes #139

### DIFF
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -91,7 +91,7 @@ class IiifController < ApplicationController
   end
 
   def load_image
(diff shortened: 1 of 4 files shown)


cbeer:
Yeah, that mirrors our authz here:

https://github.com/sul-dlss/stacks/blob/master/app/models/ability.rb#L53-L56

Tile requests can be 512x512, but thumbnails are 400x400. It's kinda silly (you could e.g. get 512-long-edge of the full image), but that's our implementation.

PR #146, opened 2017-08-03
modify the login text and add failure for auth service

mejackreed:
Closes https://github.com/sul-dlss/universalviewer/issues/32
Closes https://github.com/sul-dlss/universalviewer/issues/34


cbeer:
Could you also make the same updates to our IIIF manifests in:

https://github.com/sul-dlss/purl/blob/master/app/models/iiif_presentation_manifest.rb#L158
https://github.com/sul-dlss/purl/blob/master/app/models/iiif3_presentation_manifest.rb#L162

PR #154, opened 2017-08-25
Instead of maintaining our own ported Rails code, now lets just use the

mejackreed:
included stuff


jcoyne:
@mejackreed looks like two failing tests. One is due to calling `ActionController::HttpAuthentication::Bearer` in the test.

mejackreed:
@jcoyne yep I never got to finishing this

aaron-collier:
Replaced by #221

PR #155, opened 2017-08-25
Supports the use case where if an image is not accessible

mejackreed:
When an image is not accessible, its height/width should be scaled
to a max dimension of 400px.

Closes https://github.com/sul-dlss/sul-embed/issues/803
Closes https://github.com/sul-dlss/sul-embed/issues/808

### DIFF
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -4,6 +4,8 @@ class IiifController < ApplicationController
   before_action :load_image
   before_action :add_iiif_profile_header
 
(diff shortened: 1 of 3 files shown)


app/controllers/iiif_controller.rb
```diff
@@ -112,14 +114,28 @@ def image_info
       end
     end
 
+    ##
+    # When we are in a degraded state, we are giving back resized width/height
```
cbeer: I'd like to see a comment emphasizing how much of a terrible, implementation-specific hack this is.

PR #163, opened 2017-09-11
Move favicon.ico into the asset pipeline. Fixes #162

### LINKED ISSUE #162: Move favicon.ico to asset pipeline
```
DEPRECATION WARNING: The asset "favicon.ico" is not present in the asset pipeline.Falling back to an asset that may be in the public folder.
This behavior is deprecated and will be removed.
To bypass the asset pipeline and preserve this behavior,
use the `skip_pipeline: true` option.
 (called from _app_views_layouts_application_html_erb___1624502808018785827_70321895494920 at /Users/jcoyne/workspace/dlss/stacks/app/views/layouts/application.html.erb:14)
```

jcoyne:


cbeer:
Won't this get finger-printed? Do we need to do something to tell the browser about where it is?

jcoyne:
Yes it will be finger printed.  The layout will also have the correct fingerprint so, what's left?

cbeer:
👌 , as long as that's present.

PR #165, opened 2017-09-11
Don't load active_record, active_job, action_mailer. Fixes #164

### LINKED ISSUE #164: schema doesn't exist warning.
```
/Users/jcoyne/workspace/dlss/stacks/db/schema.rb doesn't exist yet. Run `rails db:migrate` to create it, then try again. If you do not intend to use a database, you should instead alter /Users/jcoyne/workspace/dlss/stacks/config/application.rb to limit the frameworks that will be loaded.
```

jcoyne:


cbeer:
I suspect this'll also require a change to the `Capfile` to stop it from running the database tasks?

PR #167, opened 2017-09-11
Update Rake

jcoyne:


greysteil:
@jcoyne - would you be up for automatically creating dependency update PRs like this one? I built [Dependabot](https://dependabot.com) to do that at my last company and we made it free for open-source - could be a time saver?

PR #168, opened 2017-09-11
A note about consumers of the legacy image service

jcoyne:


config/routes.rb
```diff
@@ -30,6 +30,7 @@
     get '/image/iiif/auth/:identifier/:region/:size/:rotation/:quality' => 'webauth#login_iiif', as: :auth_iiif
   end
 
+  # As of Sept 2017, the legacy service was still used by Revs and Bassi Verati/FRDA
```
cbeer: This sounds like an exclusive set, but it's likely used by other applications too.

PR #172, opened 2017-09-12
Refactoring to enable swapping image drivers

jcoyne:
- [x] Extract `IiifController#image_info`
- [x] Decouple `StacksImage#info` from knowledge about `Djatoka`


app/models/restricted_image.rb
```diff
@@ -0,0 +1,22 @@
+# The type of image that is used if a user doesn't have
+# `can? :download, stacks_image' permissions
+class RestrictedImage < StacksImage
```
cbeer: I wonder if it'd help to distinguish tile-only vs thumbnail-only images? I know we weren't necessarily doing that before, but I had trouble grokking what a `RestrictedImage` was at first.

app/models/stacks_image.rb
```diff
@@ -41,9 +66,10 @@ def druid
   end
 
   def path
-    path = super
-
-    path + '.jp2' if path
+    @path ||= begin
+                pth = PathService.for(druid, file_name)
+                pth + '.jp2' if pth
```
cbeer: Does this `jp2` suffix logic seem worth pushing into `PathService`?

app/services/auth_service.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+# Produces iiif services for authenticating
+class AuthService
+  # helper method to create a new auth service
```
cbeer: `AuthService` and `LocationService` (which I'd argue for calling `LocationAuthService`?) seem to share a lot in common, no? Maybe there's a base `IiifService` we're missing?

app/controllers/iiif_controller.rb
```diff
@@ -98,71 +102,21 @@ def current_image
...
-    info['profile'] =
-      if can? :download, current_image
-        'http://iiif.io/api/image/2/level1'
-      else
-        ['http://iiif.io/api/image/2/level1', { 'maxWidth' => 400 }]
-      end
-
+    info = current_image.info
+    info['profile'] = current_image.profile
     info['sizes'] = [{ width: 400, height: 400 }] unless current_image.maybe_downloadable?
```
cbeer: I don't remember the `maybe_downloadable?` logic, but I think this is also a candidate for pushing into `RestrictedImage`?

app/models/stacks_image.rb
```diff
@@ -1,10 +1,31 @@
...
+    RestrictedImage.new(region: region,
+                        size: size,
+                        rotation: rotation,
+                        quality: quality,
+                        format: format,
+                        id: id,
+                        file_name: file_name)
+  end
+
+  # TODO: remove tight coupling to djatoka
```
cbeer: What are you thinking about this? Maybe getting the raw IIIF JSON hash out of djatoka and then directly modifying it?

jcoyne: Yes, exactly.

app/services/auth_service.rb
jcoyne: It's a lot of boilerplate.  I don't know if it's worth extracting.

app/models/stacks_image.rb
jcoyne: I was thinking about that for a few moments., but tried to keep focused on other matters.  Is it important?

app/models/restricted_image.rb
jcoyne: To be honest, I'm not really sure what it is about either... this was just a way to keep the models from being aware of `Ability`.  It was clear that we had two classes here with lots of `if`s making it clear there was want of another class.

PR #177, opened 2017-09-13
[WIP] Extract DjatokaAdapter into a class for dynamic loading

aaron-collier:
Here is a first pass at pulling the DjatokaAdapter stuff into a class that can be called dynamically based on the the configuration, a few notes (and will add comments inline)

- Updates settings.yml to set a "driver", resolver and image class based on the driver.
- Sets the Image service resolver based on settings in an initializer
- Pulls some of the work into StacksImage so that the DjatokaImage is only used to load the image and metadata
- This is making the assumption that other image servers (RIIIF/Cantaloupe) will require a persistent resolver available.

TODO:

- Document settings.yml
- Document DjatokaImage requirements in order to create CantaloupeImage and/or RiiifImage
- Write tests and perform more complete tests against a wider range of image and metadata requests.

### DIFF
--- a/app/models/concerns/image_adapter.rb
+++ b/app/models/concerns/image_adapter.rb
@@ -1,0 +1,7 @@
+# Module for storing a configurable Image Server Resolver
+module ImageAdapter
+  # Resolver class allows catching the endpoint for the Image Server
+  class Resolver
+    class_attribute :endpoint
+  end
+end
--- a/config/initializers/image_adapter.rb
+++ b/config/initializers/image_adapter.rb
@@ -1,0 +1,1 @@
+ImageAdapter::Resolver.endpoint = Djatoka::Resolver.new(Settings.stacks.djatoka.url)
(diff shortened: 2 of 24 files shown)


app/models/concerns/image_adapter.rb
```diff
@@ -0,0 +1,9 @@
+# Common methods for projections (StacksImage, StacksMediaStream) that
+# are backed by a file on disk.
+module ImageAdapter
```
aaron-collier: This is the module used to instantiate an image resolver at startup.

config/initializers/image_adapter.rb
```diff
@@ -0,0 +1 @@
+ImageAdapter::Resolver.endpoint = Djatoka::Resolver.new(Settings.stacks.djatoka.url)
```
aaron-collier: This creates the initial resolver. I have found that an error eventually will cause this to be nil, so further testing needs to be included in case it may need to be reset at runtime.

aaron-collier:
Replaced by #183

PR #185, opened 2017-09-13
Move djatoka_url in settings to driver.image.attributes.url

aaron-collier:
This moves driver specific settings into attribute values for the image and the info services. The image and info factories will now manage setting the attributes.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,10 +1,10 @@
 stacks:
   storage_root: /stacks
-  djatoka_url: 'http://localhost:8080/resolver'
   driver: djatoka
   djatoka:
     image: DjatokaImage
     info: DjatokaInfoService
+    url: 'http://localhost:8080/resolver'
 
 purl:
   url: 'https://purl.stanford.edu/'
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -2,9 +2,12 @@ stacks:
   storage_root: /stacks
   driver: djatoka
   djatoka:
-    image: DjatokaImage
-    info: DjatokaInfoService
-    url: 'http://localhost:8080/resolver'
+    image:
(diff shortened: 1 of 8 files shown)


config/settings.yml
```diff
@@ -1,10 +1,10 @@
 stacks:
   storage_root: /stacks
-  djatoka_url: 'http://localhost:8080/resolver'
   driver: djatoka
   djatoka:
     image: DjatokaImage
     info: DjatokaInfoService
+    resolver: 'http://localhost:8080/resolver'
```
cbeer: Is `resolver` the right, cross-implementation word? It seems more like a base url to me?

aaron-collier: Actually - that's a good question that I meant to ask this morning. Are these "resolvers" in RIIIF/Cantaloupe implementations, or should this just be `uri` and then (in this case) Djatoka is using that for it's "resolver"?

aaron-collier: Changed to URL.

config/settings.yml
```diff
@@ -1,10 +1,10 @@
 stacks:
   storage_root: /stacks
-  djatoka_url: 'http://localhost:8080/resolver'
   driver: djatoka
   djatoka:
     image: DjatokaImage
     info: DjatokaInfoService
+    url: 'http://localhost:8080/resolver'
```
jcoyne: This really seems like a property of DjatokaImage.   How would you feel about:
```
   djatoka:
     image:
       class: DjatokaImage
       attributes:
          url: 'http://localhost:8080/resolver'
     info:
       class: DjatokaInfoService

```

jcoyne: Then we can have the factory set all the attributes.

aaron-collier: WFM

PR #207, opened 2017-09-14
Replaces check for downloadable and accessible on degraded images

aaron-collier:
removes generic_stanford_webauth_ability for degraded images


jcoyne:
Well done! Thanks for fixing this guys. 👏

PR #212, opened 2017-09-15
Add a factory for image server status checks

jcoyne:


app/models/djatoka_image.rb
```diff
@@ -26,6 +26,10 @@ def valid?
     image_url.present?
   end
 
+  def self.ok_check
```
cbeer: What's this doing here?

app/services/driver_factory.rb
```diff
@@ -0,0 +1,41 @@
+# @abstract Factorys that use the settings.yml
```
cbeer: typo

app/models/djatoka_image.rb
jcoyne: Yeah, cruft.

app/services/driver_factory.rb
```diff
@@ -0,0 +1,41 @@
...
+class DriverFactory
+  # Instantiate the implementation class with all passed arguments and
+  # configured attributes
+  def self.create(*args)
```
cbeer: Would `**kwargs` just work here?

config/settings.yml
```diff
@@ -8,6 +8,8 @@ stacks:
         url: 'http://localhost:8080/resolver'
     metadata:
       implementation: DjatokaMetadataService
+    check:
```
cbeer: I guess I'm unclear why the check needs to be independently configurable from the image or metadata implementation? It seems like a lot of not-entirely-straightforward code to get to (something a little better than:)

```
class DjatokaImage
def self.ok_check
  OkComputer::HttpCheck.new(Settings.stacks['djatoka'].image.attributes.url + '?rft_id=/&svc_id=info:lanl- repo/svc/ping&url_ver=Z39.88-2004')
end
```

I defer to you and @aaron-collier though. I'm not sure I have the full context for most of the recent refactoring.

jcoyne: @cbeer The idea is that when you flip the variable in settings.yml then, okcomputer does the correct checks automatically.   Presently if you flip the variable in settings, then initialization fails on okcomputer.

jcoyne: @cbeer I think you're right about the simplification, but there's still some good refactoring in here just for the base class of the factory.  I think you mentioned that in a comment on a different ticket.

jcoyne: All right, I remember why I did things this way.  Presently there's no great way of saying "what is the current image class" I guess we could make `StacksImageSourceFactory.implementation` a public method and use that in `config/initializers/okcomputer.rb`

aaron-collier: I'm inclined to agree with @cbeer on the simplicity of putting this in the implementation. But I also agree that the DriverFactory work is good refactoring toward a more robust structure for switching out the driver - it may just be a bit unnecessary since we know we're just switching directly to cantaloupe.

PR #221, opened 2017-09-19
ActionController::HttpAuthentication Refactoring

aaron-collier:
This replaces #154 with a rebased and updated version.


aaron-collier:
https://github.com/sul-dlss/stacks/blob/rebased_154/spec/features/iiif_spec.rb#L17

The current spec error is the result of the current_user in the call above looks like:
`#<User:0x007fc29d96a430 @ip_address="127.0.0.1", @anonymous_locatable_user=true>`

Causing the token auth to fails because an `anonymous_locatable_user` is not a `token_eligible_user`. Thoughts/feedback from @mejackreed here?

cbeer:
Does `Bearer` on line 17 need to be `Token` to match the upstream change?

PR #234, opened 2017-09-20
Update to the latest version of Ruby

jcoyne:


.travis.yml
```diff
@@ -2,8 +2,8 @@ notifications:
   email: false
 
 rvm:
-  - 2.3.4 # deployed
-  - 2.4.1
+  - 2.3.5
```
cbeer: can we keep testing 2.3.4 until we upgrade to 2.4.2 (or, testing 2.3.5 seems unnecessary, since I'd assume we'll take the leap right to 2.4).

PR #236, opened 2017-09-20
Remove support for invalid regions. Fixes #186

### LINKED ISSUE #186: pct:50 pct:200 are not valid region values
What is going on here: 
https://github.com/sul-dlss/stacks/blob/88a422bb6ceccc493031ccf68cf29c40ed77a3fe/spec/models/stacks_image_spec.rb#L122-L123

jcoyne:


app/models/projection.rb
```diff
@@ -21,6 +21,8 @@ def region_dimensions
     case transformation.region
     when 'full', /^pct/
       scaled_region_dimensions
+    when /^pct:/
+      raise NotImplementedError, "Percent regions are not yet supported"
```
cbeer: "Not yet" imlies we will support them?

jcoyne: we might, we just haven't yet.

Issue #241, opened 2017-09-21
[Stacks/prod] JSON::ParserError: 765: unexpected token at ''


cbeer:
## Backtrace

    line 52 of [PROJECT_ROOT]/app/services/iiif_metadata_service.rb: json
    line 33 of [PROJECT_ROOT]/app/services/iiif_metadata_service.rb: image_width
    line 50 of [PROJECT_ROOT]/app/models/stacks_image.rb: image_width

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49299/faults/34745233)

aaron-collier:
#246 - in progress

PR #246, opened 2017-09-21
Raise an exception message with JSON is invalid.

aaron-collier:
This raises an error with the HTTP response is a 200 but the JSON received is invalid. 

Addressing issue #241

### DIFF
--- a/app/services/iiif_metadata_service.rb
+++ b/app/services/iiif_metadata_service.rb
@@ -55,7 +55,7 @@ class IiifMetadataService < MetadataService
                   data['@id'] = @canonical_url
                 end
               end
-  rescue JSON::ParserError => error
-    raise Stacks::UnexpectedMetadataResponseError.new(@url, error, retrieved_json)
+  rescue JSON::ParserError => err
+    raise Stacks::UnexpectedMetadataResponseError, "There was a problem fetching #{@url}. #{err}: #{retrieved_json}"
   end
 end
--- a/app/services/iiif_metadata_service.rb
+++ b/app/services/iiif_metadata_service.rb
@@ -49,10 +49,13 @@ class IiifMetadataService < MetadataService
   end
 
   def json
+    retrieved_json = retrieve
     @json ||= begin
-                JSON.parse(retrieve).tap do |data|
+                JSON.parse(retrieved_json).tap do |data|
                   data['@id'] = @canonical_url
                 end
               end
+  rescue JSON::ParserError => error
+    raise Stacks::UnexpectedMetadataResponseError.new(@url, error, retrieved_json)
   end
 end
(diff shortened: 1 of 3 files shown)


app/services/iiif_metadata_service.rb
```diff
@@ -53,5 +53,7 @@ def json
                   data['@id'] = @canonical_url
                 end
               end
+  rescue JSON::ParserError
+    raise "There was a problem, #{@url} returned invalid JSON."
```
jcoyne: Can you provide the response in the message too? Also, how about making a custom error?  How about `Stacks::UnexpectedMetadataResponse`?

aaron-collier: For sure.

app/services/stacks/unexpected_metadata_response.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# A custom exceptions class to provide specific Stacks::Exceptions
+module Stacks
+  # An UnexpectedMetadataResponse to raise when HTTP returns bad JSON
+  class UnexpectedMetadataResponse < StandardError
+    def initialize(url = nil, message = nil)
```
jcoyne: When you call this you are passing a `ParserError` to the argument you are  calling `message`

spec/services/iiif_metadata_service_spec.rb
```diff
@@ -8,42 +8,64 @@
...
+    before do
+      allow(HTTP).to receive(:get)
+        .with('https://sul-imageserver-uat.stanford.edu/cantaloupe/iiif/2/nr%2F349%2Fct%2F7889%2Fnr349ct7889_00_0001.jp2/info.json')
+        .and_return(bad_response)
+    end
+
+    describe "#fetch" do
+      subject { service.fetch(256) }
+      it "raises Stacks::UnexpectedMetadataResponse" do
+        expect { subject }.to raise_error { |error|
```
jcoyne: change to `expect { subject }.to raise_error Stacks::UnexpectedMetadataResponse`

app/services/iiif_metadata_service.rb
```diff
@@ -49,10 +49,13 @@ def retrieve
...
   def json
+    retrieved_json = retrieve
     @json ||= begin
-                JSON.parse(retrieve).tap do |data|
+                JSON.parse(retrieved_json).tap do |data|
                   data['@id'] = @canonical_url
                 end
               end
+  rescue JSON::ParserError => error
+    raise Stacks::UnexpectedMetadataResponseError.new(@url, error, retrieved_json)
```
cbeer: I don't think we need to pass `error` as a parameter; it should be available in the exception as a `cause`, maybe?

PR #250, opened 2017-09-21
Implement etag on images from iiif

jcoyne:
I think this was overlooked. The easiest way to do this is to just
acknowledge that `/stacks` is mounted.  This has the benefit of reducing
network calls.

Also move the `valid?` check to the Projection.

### DIFF
--- a/app/models/source_image.rb
+++ b/app/models/source_image.rb
@@ -10,17 +10,11 @@ class SourceImage
     end
   end
 
-  def exist?; end
-
   def valid?; end
 
-  def etag; end
-
-  def mtime; end
-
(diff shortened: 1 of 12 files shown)


app/models/source_image.rb
```diff
@@ -10,17 +10,12 @@ def response
...
 
-  def exist?; end
-
   def valid?; end
 
-  def etag; end
-
-  def mtime; end
-
   # I think this is unnecessary in later versions https://github.com/bbatsov/rubocop/pull/3883
```
cbeer: Is this comment garbage too?

PR #258, opened 2017-09-22
Redirect to URIs when the file extension is missing.

### LINKED ISSUE #193: Should we force IIIF identifer to have .jp2
This would make it consistent with the media endpoint which has `.mp4` (or similar) in its file identifer.  This could be done with a redirect.

jcoyne:
Stacks used to allow requesting images without an extension. However
this was different than how requests for media or file downloads worked.
Now we prefer the full file name for consistency across all endpoints.

Fixes #193


mejackreed:
As a preference, I prefer the no-extension approach (maybe I'm alone in that)

PR #260, opened 2017-09-22
Avoid using the splat operator

### LINKED ISSUE #252: [Stacks/prod] ArgumentError: wrong number of arguments (given 0, expected 1..2)
## Backtrace

    line 63 of [PROJECT_ROOT]/app/controllers/application_controller.rb: token_user
    line 45 of [PROJECT_ROOT]/app/controllers/application_controller.rb: bearer_auth_user
    line 19 of [PROJECT_ROOT]/app/controllers/application_controller.rb: current_user

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49299/faults/34617984)

URI: https://stacks.stanford.edu/file/druid:xy547ss8747/Panasonic_Winter_Report2014.pdf

This only appears to happen when the user agent is Microsoft Word

jcoyne:
The token_user expect 1..2 arguments, but if the token_and_options returns
nil, then we end up sending 0, causing an error.

Fixes #252

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -42,7 +42,8 @@ class ApplicationController < ActionController::Base
   end
 
   def bearer_auth_user
-    token_user(*token_and_options(request))
+    token, options = token_and_options(request)
+    token_user(token, options)
   end
 
   def bearer_cookie


jcoyne:
See #259

PR #263, opened 2017-09-22
Projections derived from readable images are themselves readable

jcoyne:


spec/abilities/ability_spec.rb
```diff
@@ -59,6 +63,7 @@
       it { is_expected.to be_able_to(:read_metadata, image) }
       it { is_expected.to be_able_to(:read, thumbnail) }
       it { is_expected.to be_able_to(:read, square_thumbnail) }
+      it { is_expected.to be_able_to(:read, big_image) }
```
cbeer: Should we also test the stanford-only + no-download cases with this image too?

jcoyne: If you think that's best. I was just following the TDD approach (write a failing test) then fix it.

Issue #240, opened 2017-09-21
[Stacks/prod] RuntimeError: There was a problem fetching http://imageserver-prod.stanford.edu/iiif/2/sv%2F051%2Fyq%2F8509%2FBranner_00017231_0001.jp2/info.json. Server returned 403


cbeer:
## Backtrace

    line 46 of [PROJECT_ROOT]/app/services/iiif_metadata_service.rb: retrieve
    line 52 of [PROJECT_ROOT]/app/services/iiif_metadata_service.rb: json
    line 33 of [PROJECT_ROOT]/app/services/iiif_metadata_service.rb: image_width

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49299/faults/34740013)

cbeer:
We're seeing this error for files that are shelved with incorrect file permissions (usually `640`). This means our image server cannot access them. I've sent to @blalbrit a list of 111k files with this problem.

blalbrit:
argo/#977 - starting from objects impacted by this.

jcoyne:
@blalbrit one of the objects in this group is https://purl.stanford.edu/sv051yq8509.xml.  It's not dark is it?

Rights looks like this:
```xml
  <rightsMetadata>
    <access type="discover">
      <machine>
        <world/>
      </machine>
    </access>
    <access type="read">
      <machine>
        <world/>
      </machine>
    </access>
    <use>
      <human type="useAndReproduction">Theses courtesy of Stanford University Libraries.  If you have questions, please contact the Branner Earth Science Library &amp;amp; Map Collections at brannerlibrary@stanford.edu.</human>
    </use>
  </rightsMetadata>
```

I'm not sure how sul-dlss/argo#977 applies.

mejackreed:
Is perhaps `xp490xd1371` also part of this? I just went to the DRMC and found many items that were returning 403's incorrectly.

jcoyne:
Looks like Honeybadger is lumping the 500 responses with the 400 responses.  I'm seeing this in the imageserver-c log for one occurance:

```
09:31:59.651 [ajp-apr-8009-exec-4] ^[[1;31mERROR^[[0;39m ^[[36mo.r.C.WebApplication^[[0;39m - Exception or error caught in server resource
org.restlet.resource.ResourceException: Internal Server Error (500) - The server encountered an unexpected condition which prevented it from fulfilling the request
...
[...]

PR #284, opened 2017-09-26
Switch exist? to readable? for StacksFiles

aaron-collier:
Addresses #240


jcoyne:
@aaron-collier I talked with Chris and we would need to switch the check to look for world readable.  This is because the imageserver runs as "tomcat" and stacks runs as the "stacks" user.

aaron-collier:
Perhaps: https://ruby-doc.org/core-1.9.3/File/Stat.html#method-i-mode

PR #303, opened 2017-10-05
Sanitize utf-8 in requests. Fixes #302

### LINKED ISSUE #302: [Stacks/dev] ArgumentError: invalid byte sequence in UTF-8
## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49299/faults/34397275)

jcoyne:


jcoyne:
Ref: https://github.com/rails/rails/issues/30811

aaron-collier:
Should we have any tests here to make sure this doesn't break anything? I'm assuming it's clean and this ultimately helps, just want to make sure we're good.

jcoyne:
@aaron-collier how would you suggest I test this?

aaron-collier:
@jcoyne - Good question, I'm assuming that adding this effects some change to a problem we saw before that could be verified? Some description to this PR would help as it's been a while ;-)

jcoyne:
No longer an issue since we no longer use Sprockets.

PR #321, opened 2018-01-24
Redirect thumbnail requests for restricted images to the degraded ide…

cbeer:
…ntifier so IIIF clients invoke the auth workflow

### DIFF
--- a/app/controllers/iiif_controller.rb
+++ b/app/controllers/iiif_controller.rb
@@ -37,7 +37,7 @@ class IiifController < ApplicationController
 
     return unless stale?(cache_headers_metadata)
 
-    if degraded? && !degraded_identifier?
+    if degraded?
       redirect_to iiif_metadata_url(identifier: degraded_identifier)
       return
     end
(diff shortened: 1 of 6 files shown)


app/controllers/iiif_controller.rb
```diff
@@ -167,11 +167,14 @@ def add_iiif_profile_header
...
-  # We consider an image to be degraded if the user isn't current able to download it, but if they
+  # We consider an image to be degraded if the user isn't currently able to download it, but if they
   # login as a stanford user, they will be able to.
   def degraded?
+    return false if degraded_identifier?
+
     !can?(:access, current_image) && current_image.accessable_by?(stanford_generic_user) ||
-      !can?(:download, current_image) && current_image.readable_by?(stanford_generic_user)
+      !can?(:download, current_image) && current_image.readable_by?(stanford_generic_user) ||
+      !can?(:access, current_image) && can?(:read, Projection.thumbnail(current_image))
```
jkeck: I'm having trouble reasoning about if `!can?(:access, current_image)` does anything given the same thing two lines above.  Does ruby automatically group the things on either side of the `&&` before executing the `||` (which is the behavior I think we're depending on here).

After reading it over carefully, I think I grok what's going on, so maybe a comment here would be helpful if refactoring this is too cumbersome.

cbeer: Yes, in ruby, `&&` has higher precedence than `||`.

cbeer: And, sure, commented.

PR #335, opened 2018-03-06
Set the X-Sendfile header for mod_sendfile

### LINKED ISSUE #334: Use X-sendfile
https://github.com/sul-dlss/stacks/blob/master/config/environments/production.rb#L34

jcoyne:
This allows big files to be delivered by Apache rather than using a
Rails server process (which is more resource intensive)

Fixes #334

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -31,7 +31,7 @@ Rails.application.configure do
   # config.action_controller.asset_host = 'http://assets.example.com'
 
   # Specifies the header that your server uses for sending files.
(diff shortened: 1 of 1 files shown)


ndushay:
👍

mejackreed:
This isn't an objection to this, but just something keeps popping up in my mind that we didn't do this for a specific reason. Perhaps media or IIIF-auth? I can't remember. I don't think this should be blocked, but just monitored for these use cases.

jkeck:
Nothing I can remember specifically from the media work FWIW.

PR #336, opened 2018-03-06
Remove unused dependencies

jcoyne:


mejackreed:
It seems like `application.js` is requiring jquery https://github.com/sul-dlss/stacks/blob/master/app/assets/javascripts/application.js#L13

jcoyne:
@mejackreed Thank you. I've removed that now too.

PR #350, opened 2018-04-03
default use of Bearer for authorization token in auth scheme, for III…

mejackreed:
…F Auth compliance

Fixes https://github.com/sul-dlss/sul-embed/issues/827
Fixes https://github.com/sul-dlss/sul-embed/issues/856

This change fixes a regression introduced in https://github.com/sul-dlss/stacks/pull/221 . The regression switched usage of the authorization header prefix from `Bearer` to `Token`. `Bearer` is required via the [IIIF Auth spec](http://iiif.io/api/auth/1.0/#the-json-access-token-response) and we were not honoring it anymore.

This change should be backwards compatible now accepting both.

### DIFF
--- a/app/controllers/iiif_token_controller.rb
+++ b/app/controllers/iiif_token_controller.rb
@@ -8,7 +8,7 @@ class IiifTokenController < ApplicationController
     @message = if token
                  {
                    accessToken: token,
-                   tokenType: 'Token',
+                   tokenType: 'Bearer',
                    expiresIn: 3600
                  }
                else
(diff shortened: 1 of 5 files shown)


mejackreed:
I've tested this on UAT with UV

app/controllers/iiif_token_controller.rb
```diff
@@ -8,7 +8,7 @@ def create
     @message = if token
                  {
                    accessToken: token,
-                   tokenType: 'Token',
+                   tokenType: 'Bearer',
```
cbeer: I think `tokenType` is gone in IIIF Auth 1.0; I'm not sure if we had backwards compatibility concerns that led us to keep it as-is?

mejackreed: Yeah I hadn't realized that. /shrug should we remove it?

PR #374, opened 2018-06-21
Add support for shibboleth groups

cbeer:


jkeck:
Ignoring code climate as we'll only be able to address those issues once we're fully deployed on shibboleth.

PR #519, opened 2019-08-05
Update SU footer

mejackreed:


cbeer:
This has the same link decoration on the SUL logo as https://github.com/sul-dlss/purl/pull/340 (and perhaps the same solution.). The SU footer has some funny hover stuff going on, but it's pre-existing, so 🤷‍♂

mejackreed:
Thanks I've made those changes here now.

PR #526, opened 2019-09-05
Refactor users and rights to allow users that should have access to multiple location to access the location-restricted material

### LINKED ISSUE #473: stacks can't recognize the same IP address in more than one user location
When a user's static IP addresses is specified in more than one location, as seen in [shared_configs](https://github.com/sul-dlss/shared_configs/blob/sul-stacks-prod/config/settings/production.yml#L47-L65), the application is unaware of more than one of the locations. 

Looking at [code from the `ApprovedLocation` class](https://github.com/sul-dlss/stacks/blob/216ccde5b655c88f663e1913061e648c15d04be8/app/models/approved_location.rb#L25-L27) we see something like:
```ruby
Settings.user.locations.find { |_,ip_addresses| ip_addresses.include?(ip_address1) }
 => [:location1, [ip_address1, ip_address2]] 
```
even if `:location2` also includes `ip_address1`

We need the ability for stacks to recognize multiple approved locations for the same ip address.

cbeer:
Fixes #473

### DIFF
--- a/app/models/approved_location.rb
+++ b/app/models/approved_location.rb
@@ -9,10 +9,12 @@ class ApprovedLocation
     @locatable = locatable
   end
 
-  def to_s
-    return '' unless location_for_ip
+  def locations
+    return [] unless locatable.try(:ip_address)
 
-    location_for_ip[0].to_s
+    location_configuration.select do |_, ip_addresses|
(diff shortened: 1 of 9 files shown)


hannahfrost:
thank you!

PR #561, opened 2020-03-10
Filter the sizes that are returned from an info.json

mejackreed:
Part of https://github.com/sul-dlss/sul-embed/issues/739

Our image server isn't able to correctly process large images. These seem to approach 250,000,000 pixels.

This sets a maximum pixel allowance of 200,000,000 as a general threshold.

Some data about download requests here: https://docs.google.com/spreadsheets/d/1aKeievR9keJVS0TKGq70rPAY7I4CWKT7DpWZsYiLURI/edit#gid=0

### DIFF
--- a/app/services/iiif_info_service.rb
+++ b/app/services/iiif_info_service.rb
@@ -3,6 +3,7 @@
 # Produces iiif info.json responses
 class IiifInfoService
   ASPECT_EXAGGERATION = 40.0
+  MAX_SIZE_ALLOWED = 200_000_000
 
   # A helper method to instantiate and produce the json
   # @param current_image [StacksImage,RestrictedImage] the image to get the information for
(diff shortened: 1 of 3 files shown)


lib/tasks/test_images.rake
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+desc 'Test image downloads for large pixel size images'
```
mejackreed: I'm happy to 🔥 this commit if we want. I found this useful to quickly gather some data.

app/services/iiif_info_service.rb
```diff
@@ -36,6 +38,15 @@ def info
     end
   end
 
+  ##
+  # Trims the provided sizes so that known large undownloadable sizes are not
+  # returned
+  def sizes(sizes)
+    sizes.to_a.filter do |size|
```
jkeck: Is there something besides an array that will come through as the `sizes` param?

mejackreed: I think in reality maybe no, but in a test case it could be `nil`

jkeck: An alternative approach might be to return an empty array in that case that sizes is nil.  Realistically it doesn't make a huge difference in our case but might avoid us needing to cast nils to an arrays (as well as save a very small mount of overhead avoiding casting arrays to arrays).

jkeck: But thinking realistically, I'm pretty sure these are all cached like crazy, and we're likely never to see the scale where this sort of optimization is likely to matter.  This code is totally understandable, so happy to :shipit:.

PR #564, opened 2020-03-12
Tell CodeClimate to use a newer version of Rubocop

jkeck:


jkeck:
FWIW, CC appears to use the config from master.  Once this is merged future PRs will start running in CC again.

PR #568, opened 2020-03-16
Don't treat blank/empty remote users from the request as a…

jkeck:
…"Webauth" user.


app/controllers/application_controller.rb
```diff
@@ -21,7 +21,7 @@ def current_user
                         bearer_auth_user
                       elsif has_bearer_cookie?
                         bearer_cookie_user
-                      elsif session[:remote_user] || request.remote_user
+                      elsif session[:remote_user] || request.remote_user.present?
```
jkeck: Not sure if this is related to the server upgrades but it looks like non-authenticated users are now returning an empty string (instead what I have to assume was `nil`)

PR #624, opened 2020-07-14
Advertise content range requests for files

cbeer:


mejackreed:
This would also enable support for Cloud optimized GeoTiffs I think. https://github.com/cogeotiff/cog-spec/blob/master/spec.md

app/controllers/file_controller.rb
```diff
@@ -12,6 +12,8 @@ def show
 
     authorize! :read, current_file
     expires_in 10.minutes
+    response.headers['Accept-Ranges'] = 'bytes'
+    response.headers['Content-Length'] = current_file.content_length
```
jkeck, line 16: I'm not super familiar w/ the range spec, but will it be an issue to advertise that we accept byte ranges if we don't know the content length?

cbeer: I don't know about an issue, but it's certainly harder for clients. In this case,  we'll always know the content length because we're grabbing it from the file we're about to serve up. It'd be harder to do it for e.g. IIIF requests where we're just streaming through what the image server is pushing out.

PR #641, opened 2020-08-25
Support IIIF authentication  services for controlled digital lending

cbeer:
This PR adds support for controlled digital lending  using the IIIF Auth  API. Items are "checked out" using the IIIF login auth pattern, forwards the user to sul-requests to perform the Symphony interactions, and returns to stacks with a JWT token that "authorizes" the viewer.

The basic version should work with any IIIF-capable viewer, but additional functionality will need to come from a custom viewer plugin.

### DIFF
--- a/app/controllers/cdl_controller.rb
+++ b/app/controllers/cdl_controller.rb
@@ -1,0 +1,125 @@
+# frozen_string_literal: true
+
+##
+# Controlled digital lending endpoint (also protected in production by shibboleth)
+class CdlController < ApplicationController
+  before_action do
+    raise CanCan::AccessDenied, 'Unable to authenticate' unless current_user
+  end
+  skip_forgery_protection only: [:show, :show_options]
+
+  before_action :write_auth_session_info, only: [:create]
+  before_action :validate_token, only: [:delete]
+
+  # Render some information about an active CDL token so the viewer can display e.g.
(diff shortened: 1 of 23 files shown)


app/controllers/cdl_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+  end
+
+  def delete_success
+    respond_to do |format|
+      format.html { render html: '<html><script>window.close();</script></html>'.html_safe }
+      format.js { render js: 'window.close();' }
+    end
+  end
+
+  def renew
```
cbeer, line 80: Should we test these?

app/controllers/cdl_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+  # As with #create, we bounce the user to requests to handle the Symphony interaction,
+  # and after they come back we'll clean up cookies + tokens on this end.
+  def delete
+    token = existing_payload[:token]
+
+    checkin_params = { token: token, return_to: cdl_checkin_success_iiif_auth_api_url(params[:id]) }
+    redirect_to "#{Settings.cdl.url}/checkin?#{checkin_params.to_param}"
+  end
+
+  def delete_success
```
cbeer, line 73: Comment about why we don't the delete token from the cookie here?

PR #643, opened 2020-08-25
Remove indirection

cbeer:
Potential future work:
- [ ] rename  `id` to `druid`
- [ ] consider passing around StacksFile objects instead of initializing them in multiple models


app/models/stacks_rights.rb
```diff
@@ -2,7 +2,14 @@
 
 ##
 # RightsMetadata interpretation
-module StacksRights
+class StacksRights
```
jcoyne, line 5: 👍

mejackreed (approved):

PR #734, opened 2021-01-06
Extract barcode extraction to the Purl class...

cbeer:
and support `otherId[@name='barcode']` as a potential source of a barcode


mejackreed (approved):
This works for me. THanks for adding the helpful tests.

PR #736, opened 2021-01-11
Update dependencies

cbeer:
CI wants to downgrade zipline for some reason...


mejackreed:
Maybe it needs a release still? https://github.com/sul-dlss/stacks/pull/733

cbeer:
Yeah, this is without the Rails 6.1 update. who knows when we can expect that (but we do want the nokogiri bump)

PR #888, opened 2022-06-17
Update to Rails 7

cbeer:


bin/setup
```diff
@@ -13,16 +13,16 @@ FileUtils.chdir APP_ROOT do
...
   # Install JavaScript dependencies
   system! 'bin/yarn'
 
   puts "\n== Removing old logs and tempfiles =="
-  system! 'bin/rails log:clear tmp:clear'
+  system! "bin/rails log:clear tmp:clear"
 
   puts "\n== Restarting application server =="
-  system! 'bin/rails restart'
+  system! "bin/rails restart"
```
thatbudakguy, line 27: did we mean to change all these quotes? i think we usually prefer single quotes on other projects. maybe just needs a rubocop rule?

config/application.rb
```diff
@@ -10,9 +10,12 @@
 # require "active_storage/engine"
 require "action_controller/railtie"
 # require "action_mailer/railtie"
+# require "action_mailbox/engine"
+# require "action_text/engine"
 require "action_view/railtie"
 # require "action_cable/engine"
 require "sprockets/railtie"
+require "sprockets/railtie"
```
thatbudakguy, line 18: seems weird to do this twice

bin/setup
cbeer: If rubocop's not yelling and we're not changing it ourselves, keeping this as-generated is nice 🤷‍♂️

PR #925, opened 2022-10-17
Allow slashes in filenames

mjgiarlo:
Connects to https://github.com/sul-dlss/dor-services-app/issues/4256

This commit supports work to support hierarchy in filenames throughout SDR. Doing this to support analysis work.


config/routes.rb
```diff
@@ -3,16 +3,30 @@
 
   get '/object/:id' => 'object#show', as: :object
 
-  constraints id: druid_regex, file_name: %r{[^/]+} do
-    get '/file/:id/:file_name' => 'file#show', as: :file
-    options '/file/:id/:file_name', to: 'file#options'
-    get '/file/app/:id/:file_name' => 'webauth#login_file'
-    get '/file/auth/:id/:file_name' => 'webauth#login_file', as: :auth_file
+  constraints id: druid_regex do
+    if Settings.features.allow_slashes_in_filenames
```
jcoyne, line 7: why do we need this conditional? can we just set these routes as required?

jcoyne (changes requested):

config/routes.rb
mjgiarlo: @jcoyne I'd be fine with that. Initially, one of the Access devs (maybe Chris?) asked if the feature could be behind a feature flag. LMK how you'd like to proceed.

jcoyne: @cbeer why do you want a feature flag here?  Can't we just revert if it doesn't work?  Otherwise we need additional testing so that we hit both paths, right?

jcoyne: @cbeer say something if you think this is required. Otherwise I think we should remove it to minimize complexity and avoid having to clean this up in the future.

mjgiarlo: @jcoyne Removed in a follow-up commit.

jcoyne (approved):

jcoyne (changes requested):
Do we need to add some requests tests that hit the FilesController and show that it can deal with files that have slashes?

spec/requests/file_spec.rb
```diff
@@ -2,17 +2,45 @@
...
+      allow(File).to receive(:world_readable?).and_return(nil)
+      allow_any_instance_of(FileController).to receive(:send_file).with(stacks_file.path, disposition: :inline)
+      allow(Purl).to receive(:public_xml).and_return(world_rights)
+    end
+
+    it 'succeeds' do
+      get '/file/xf680rd3068/path/to/xf680rd3068_1.jp2'
+      expect(response).to be_successful
+    end
+  end
```
mjgiarlo, line 45: Cribbed authZ code from `spec/requests/file_auth_request_spec.rb`.

mjgiarlo:
@jcoyne 💬 
> Do we need to add some requests tests that hit the FilesController and show that it can deal with files that have slashes?

Added a request test.

spec/requests/file_spec.rb
```diff
@@ -2,17 +2,45 @@
...
+                <world/>
+              </machine>
+            </access>
+          </rightsMetadata>
+        </publicObject>
+      EOF
+    end
+
+    before do
+      allow(File).to receive(:world_readable?).and_return(nil)
```
jcoyne, line 36: Is this needed?  Where do we call `File.world_readable?`?

jcoyne, line 36: Why are we returning files that aren't readable? Is this stub necessary?

mjgiarlo: @jcoyne Good eye! It was copypasta from the file auth request spec. Removed it and all is well.

jcoyne (approved):

PR #928, opened 2022-11-04
Keep redis gem on v4

corylown:
This will unblock dependency updates and deploys until Redis is updated and we can update Sidekiq to version 7. See: https://github.com/sul-dlss/operations-tasks/issues/3210#issuecomment-1302508219


corylown:
May not be needed

PR #1003, opened 2023-11-02
handling dimensions that are too large

hudajkhan:
Closes https://github.com/sul-dlss/sul-embed/issues/1604 

What this pull request does:

- For requests that ask for an explicit width and height combination that is larger than the original image size, this code uses the original image size in the IIIF image request instead.  
- Code already existed to handle restricted images (which relate to user permissions).  That code has been updated and repurposed to check for this situation and use the original image size. 
- The type of image class seems to be different between the restricted image use case and these full size image requests.  The former has a different method for calculating image size, so the code checks which method is available and uses the appropriate one.

### DIFF
--- a/spec/models/projection_spec.rb
+++ b/spec/models/projection_spec.rb
@@ -20,7 +20,7 @@ RSpec.describe Projection do
     end
   end
 
-  describe '#tile_dimensions' do
+  describe '#tiledimensions' do
     subject { instance.send(:tile_dimensions) }
 
     context "for an unrestricted image" do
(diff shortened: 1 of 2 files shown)


spec/models/projection_spec.rb
```diff
@@ -20,7 +20,7 @@
     end
   end
 
-  describe '#tile_dimensions' do
+  describe '#tiledimensions' do
```
jcoyne, line 23: I think this was in error.

jcoyne (changes requested):

spec/models/projection_spec.rb
hudajkhan: yes

jcoyne (approved):

PR #1049, opened 2023-11-21
send some basic access restriction info along with successful auth check response

jmartin-sul:
useful for indicating restrictions that would apply for other users even if current user is logged in

required by https://github.com/sul-dlss/sul-embed/pull/1763 (but should be backwards compatible with sul-embed before that change).

conencts with https://github.com/sul-dlss/sul-embed/issues/1531


jmartin-sul:
unit tests pass in local dev environment for me, will look to see if CI is running things differently, may have to get a second set of eyes if nothing jumps out

.github/workflows/ruby.yml
```diff
@@ -31,4 +31,4 @@ jobs:
     - name: Run tests
       run: bin/rake
     - name: Run tests (with cocina)
-      run: SETTINGS__FEATURES__COCINA=true bin/rake spec
+      run: bin/rake spec
```
jcoyne, line 34: Why is this setting being removed?

jcoyne (changes requested):

.github/workflows/ruby.yml
jmartin-sul: d'oh, i was not looking closely enough at the CI steps, and missed that CI is running `bin/rake spec` both with and without this setting being true (i mistakenly thought we were only running specs with this override, and was trying to make default `rspec` behave on laptop the way i thought CI tests were behaving, hence the `test.yml` change, since i'd gotten a couple failures in CI that i didn't initially get locally, and realized that i had to run with `SETTINGS__FEATURES__COCINA=true` to get those same failures).

anyway, will undo this change, thanks for pointing this out.

jmartin-sul:
@jcoyne i removed the offending commit, ready for another look

spec/controllers/media_controller_spec.rb
```diff
@@ -106,6 +115,17 @@
         expect(body['status']).to eq 'success'
         expect(body['token']).to eq 'sekret-token'
       end
+
+      it 'still returns info about applicable access restrictions' do
```
jcoyne, line 119: I'm not sure I understand "still" here?  Can we remove that?

spec/requests/media_auth_request_spec.rb
```diff
@@ -68,6 +68,11 @@
       expect(response.parsed_body['status']).to eq 'success'
       expect(response.parsed_body['token']).to match(/^[%a-zA-Z0-9]+/)
     end
+
+    it 'still indicates that the object is stanford restricted' do
```
jcoyne, line 72: here as well?

jcoyne (changes requested):

spec/controllers/media_controller_spec.rb
jmartin-sul: ah, yeah, was thinking relative to the old behavior where access restriction info was no longer returned once the user is successfully logged in.  but i can see how that's confusing, removed.

jcoyne (approved):

PR #1053, opened 2023-12-01
Run container in development env

mjgiarlo:
We do this so Stacks can serve files (caption files, media files, etc.) directly without needing e.g. Wowza or passenger

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,21 +8,11 @@ RUN apk add --update --no-cache  \
 
 WORKDIR /app
 
-ENV RAILS_ENV="production" \
-    BUNDLER_WITHOUT="development test" \
-    RAILS_SERVE_STATIC_FILES="true"
-
 COPY Gemfile Gemfile.lock ./
 
 RUN bundle install
 
 COPY . .
 
-# Precompile bootsnap code for faster boot times
-RUN bundle exec bootsnap precompile --gemfile app/ lib/
-
-# Precompiling assets for production without requiring secret RAILS_MASTER_KEY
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -25,4 +21,4 @@ RUN bundle exec bootsnap precompile --gemfile app/ lib/
 RUN SECRET_KEY_BASE_DUMMY=1 bundle exec rails assets:precompile
 
 EXPOSE 3000
-CMD ["./bin/rails", "server"]
\ No newline at end of file
+CMD ["./bin/rails", "server", "--binding=0.0.0.0"]
```
mjgiarlo, line 24: The explicit host binding is needed when running the server in the `development` Rails env.

mjgiarlo:
@jcoyne 💬 
> Can you explain more about why we want to change the env to developement? It doesn't seem like we'd use this for development (with code reloading and such) since we aren't sharing the rails directory as a volume.

So that Puma serves files via `send_file` (in `FileController#show`) without needing Passenger, Nginx, or Apache. We **could** run Stacks in the `production` Rails env and serve files via `send_file` by setting `config.action_dispatch.x_sendfile_header` to a ternary that  sets the value to `nil` in containerland and `X-Sendfile` in deployed environments (so Passenger/Apache serves files)... but it looks odd to me to change the production config for what is a change we only use when doing local dev. Thus, using the `development` Rails env. This is also consistent with making the smallest change possible to support local dev improvements.

jcoyne:
🤷  I think the ternary in config/production.rb is my preference.

jcoyne (approved):

PR #1068, opened 2023-12-06
Add cors headers for file show method

### LINKED ISSUE #1066: Correctly set Access-Control-Allow-Origin when serving Stanford Only files
In order for PDF and 3D Objects to be viewable in sul_embed JavaScript needs to fetch the file (PDF or GLB) from Stacks. This works fine in cases where the file is public. But if the file is *Stanford Only* the fetch request needs to [include credentials](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included) (assuming they are logged in) or else the request will be denied.

Unfortunately, it's not possible to send credentials unless Stacks sends `Access-Control-Allow-Origin: embed.stanford.edu` in its response instead of `Access-Control-Allow-Origin: embed.stanford.edu: *`. From the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included):

> Note: Access-Control-Allow-Origin is prohibited from using a wildcard for requests with credentials: 'include'. In such cases, the exact origin must be provided; even if you are using a CORS unblocker extension, the requests will still fail.

It looks to me like the proposed work in #1063 would send the correct CORS header when a Stanford Only file is being sent?

See also: https://github.com/sul-dlss/sul-embed/issues/1642

peetucket:
Basically adds the CORs header to any request to show a file (not just media auth requests, as is current).

Fixes #1066

### DIFF
--- a/app/controllers/file_controller.rb
+++ b/app/controllers/file_controller.rb
@@ -7,6 +7,8 @@ class FileController < ApplicationController
     render plain: 'File not found', status: :not_found
   end
 
+  before_action :set_cors_headers, only: [:show], if: proc { current_file.stacks_rights.stanford_restricted? }
+
   # rubocop:disable Metrics/AbcSize
   def show
     return unless stale?(**cache_headers)
--- a/spec/controllers/file_controller_spec.rb
+++ b/spec/controllers/file_controller_spec.rb
@@ -11,6 +11,30 @@ RSpec.describe FileController do
     Factories.cocina_with_file
   end
 
+  let(:stanford_json) do
+    {
+      'structural' => {
+        'contains' => [
+          {
+            'structural' => {
+              'contains' => [
+                {
+                  'filename' => 'xf680rd3068_1.jp2',
+                  'access' => {
+                    'view' => 'stanford',
+                    'download' => 'stanford'
+                  }
+                }
+              ]
+            }
+          }
+        ]
+      }
+    }
+  end
+
+  let(:file) { StacksFile.new(id: druid, file_name: 'xf680rd3068_1.jp2') }
+
   describe '#show' do
     let(:druid) { 'nr349ct7889' }
 
@@ -20,6 +44,7 @@ RSpec.describe FileController do
     it 'sends the file to the user' do
       expect(controller).to receive(:send_file).with(path, filename: 'image.jp2', disposition: :inline).and_call_original
       subject
+      expect(response.headers.to_h).to include 'Access-Control-Allow-Origin' => '*'
     end
 
     context 'when file is not in a content addressable path' do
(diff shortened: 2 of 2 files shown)


peetucket:
@edsu  this would add the CORs header to any request for a PDF (or any other file download for that matter)

Perhaps would allow the current PDF.js based viewer to work if the user is authed?  Still doesn't deal with any messaging if they are not authed of course (as that is a different set of work).

edsu:
I think we only want to set `Access-Control-Allow-Origin: embed.stanford.edu` in cases where the file is Stanford Only? If we don't then any JavaScript running elsewhere won't be able to access any Stacks files?

Admittedly there may not be many places that do this. I know that @peterchanws has put web archive links into items that point to replayweb.page that then fetch data from stacks. For example https://purl.stanford.edu/ww180gg9763 which links to:

https://replayweb.page/?source=https://stacks.stanford.edu/file/druid:ww180gg9763/the_quaranzine-20210609.wacz

Stanford University Press is doing something similar for some of its items, which use JavaScript to replay web archives stored in stacks: https://archive.supdigital.org/

peetucket:
> I think we only want to set `Access-Control-Allow-Origin: embed.stanford.edu` in cases where the file is Stanford Only? If we don't then any JavaScript running elsewhere won't be able to access any Stacks files? Those will need `Access-Control-Allow-Origin: *` as it currently is.

Possibly.  Adjusted PR for this situation.  Did not test anything.

peetucket:
The new test passes locally for me...not sure why it fails in CI.

app/controllers/file_controller.rb
```diff
@@ -7,7 +7,7 @@ class FileController < ApplicationController
     render plain: 'File not found', status: :not_found
   end
 
-  before_action :set_cors_headers, only: [:show]
+  before_action :set_cors_headers, only: [:show], if: proc { current_file.stacks_rights.stanford_restricted? }
```
edsu, line 10: well that looks easy -- nicely done!

spec/controllers/file_controller_spec.rb
```diff
@@ -58,6 +80,19 @@
...
+        allow(Purl).to receive(:public_json).and_return(stanford_json)
+      end
+
+      it 'sends CORs header' do
+        subject
+        expect(response.headers.to_h).to include 'Access-Control-Allow-Origin' => 'https://embed.stanford.edu',
+                                                 'Access-Control-Allow-Credentials' => 'true'
+      end
+    end
+
```
edsu, line 95: Is it worth adding a test for when the file is not Stanford Only to make sure the header is `*`?

edsu: Sure, I can take a look.

edsu: I see that CI is running the tests like this, which fails for me locally (unlike a regular rspec run):

    SETTINGS__FEATURES__COCINA=true bin/rake spec

peetucket: Ahh, yes.  good catch.  likely an issue with the mocking in the test not accounting for cocina rights computation

peetucket: Fixed. Had a problem in my mocked cocina json (it wasn't internally consistent for stanford only).

peetucket:
I don't think any of this is needed, because it turns out we set the CORs header globally in stacks and allow any domain, see https://github.com/sul-dlss/stacks/blob/main/app/controllers/application_controller.rb#L11-L19

As to why it isn't respected in some cases, it may have been due to this recent change: https://github.com/sul-dlss/stacks/pull/1096

Though that may not fully explain.  Regardless, this code is likely duplicative of what already exists.

edsu:
The reason why it is needed is that when JavaScript does a cross-origin fetch with credentials the wildcard CORS header will not work.

peetucket:
Should this PR be reopened then?

edsu:
I guess it could be if we want to push #1944 forward.

jcoyne:
I believe this was supeseeded by https://github.com/sul-dlss/stacks/pull/1100

PR #1081, opened 2023-12-11
Add a IIIF v2 token service

jcoyne:


app/controllers/iiif/auth/v2/token_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+      # Creates tokens for IIIF auth v2
+      # https://iiif.io/api/auth/2.0/#access-token-service
+      class TokenController < ApplicationController
+        skip_forgery_protection
+
+        # Returns an HTML response that posts back to the parent window
+        # See {https://iiif.io/api/auth/2.0/#workflow-from-the-browser-client-perspective}
+        def create
+          params.require(:origin)
+          params.require(:messageId)
```
mjgiarlo, line 16: Could you do this in a single line, by passing both symbols in an array? I don't think it's a big deal either way, perhaps a bit less noisy.

app/controllers/iiif/auth/v2/token_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        # An authenticated user can retrieve a token if they are logged in with webauth,
+        # or are accessing material from a location-specific kiosk.
+        # Other anonymous users are not eligible.
+        def token_eligible_user?
+          current_user.webauth_user? || current_user.location?
+        end
+
+        def mint_bearer_token
+          "#{ActionController::HttpAuthentication::Token::TOKEN_KEY}#{current_user.token.to_s.inspect}"
```
mjgiarlo, line 60: Are the `#to_s` and `#inspect` needed here?

spec/requests/iiif/auth/v2/tokens_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    context 'when there is an error' do
+      let(:user) { User.new id: 'xyz', webauth_user: false }
+
+      it 'posts the response' do
+        expect(response.response_code).to eq 200
+        expect(response.headers['X-Frame-Options']).to eq ''
+        payload = JSON.parse(/({.*\})/.match(response.body)[1])
+        expect(payload).to include(
+          { "@context" => "http://iiif.io/api/auth/2/context.json",
```
edsu, line 38: Is it worth ensuring that `accessToken` is not included in the payload here? Or does this do that already?

edsu (approved):
I'm not up on all the details of how this is supposed to work, but this PR seems additive for stacks, and unlikely to cause any problems if tests are otherwise passing.

app/controllers/iiif/auth/v2/token_controller.rb
jcoyne: I'm not really sure. I'm copying from https://github.com/rails/rails/blob/eb32f1ec40d2a355dfb3f4c4ef358917de32e26c/actionpack/lib/action_controller/metal/http_authentication.rb#L536

spec/requests/iiif/auth/v2/tokens_spec.rb
jcoyne: I've switched the assertion to an eq, because there shouldn't be anything else here.

app/controllers/iiif/auth/v2/token_controller.rb
jcoyne: Cool. I didn't know about that way.

PR #1086, opened 2023-12-14
Track file, object, and IIIF image downloads

### LINKED ISSUE #1048: Send download events to the SDR metrics service
See discussion on https://github.com/sul-dlss/sul-embed/issues/1683.

We want to track both "download" scenarios:
* `/object/:id`, which is `ObjectsController#show`, that downloads a .zip of all the files
* `/file/:id/*file_name`, which is `FilesController#show`, that serves a single file

We will need to re-implement some pieces of Ahoy; namely:
- generating unique visit and visitor tokens (UUIDs)
- storing the tokens as cookies and expiring them after some time passes
- detecting some stats about the user from their user agent (Ahoy uses the `device_detector` gem for this)

We should end up with a new service that is capable of POSTing download events to the SDR metrics API in [Ahoy's format](https://github.com/ankane/ahoy#api-spec).

thatbudakguy:
Closes #1048

This implementation is very permissive about what counts as a download.

- Any time a file is served, that's one download of that file (e.g. loading a pdf in the pdf viewer) for that druid
- Any time IIIF images are loaded in a viewer, it counts as downloading that image for that object's druid
- Any time an entire object is downloaded as a .zip, that's one download of the druid (no file specified)

This means that for some types of content (IIIF images, pdfs) viewing them also automatically downloads them at least once.

It's possible to be more selective about what counts as a download by only tracking events where the user explicitly sent the `?download=true` parameter, which we do when you click a download link in sul-embed. However, this would mean that we likely would miss out on many events coming from outside the viewer, like someone linking directly to an image from stacks on their website, because they won't use the `?download=true` parameter to do that.

Once this is merged, we should merge https://github.com/sul-dlss/sul-embed/pull/2028 to avoid double-counting downloads, and https://github.com/sul-dlss/purl/pull/917 to display download counts for IIIF objects (since we can now track them).

Finally, https://github.com/sul-dlss/shared_configs/pull/2119 will turn it on.


app/controllers/concerns/metrics_concern.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# Methods for logging usage metrics based on requests for files
+module MetricsConcern
+  def track_download(druid, file: nil)
```
jcoyne, line 5: Can this be made Async so that it doesn't block downloads?

jcoyne (changes requested):

app/controllers/concerns/metrics_concern.rb
thatbudakguy: For sure, this seems like a great use case for async. Should I just use the `async` gem? Never done it before in ruby.

jcoyne: I was thinking ActiveJob.

thatbudakguy: Updated to use a job

jcoyne (approved):

PR #1089, opened 2023-12-15
allow probe service URLs to work even if druid: prefix is not provided in URI

peetucket:
The IIIF manifest currently provides stacks URIs without a druid: prefix, this breaks our parsing, which expects it.

e.g. https://purl.stanford.edu/bh502xm3351/iiif3/manifest

has `https://stacks.stanford.edu/file/bh502xm3351/CAS_bh502xm3351.pdf` in the `items` node.  

This is a valid URI that provides the PDF, but if you call the auth service with it, you get a 404:

https://stacks.stanford.edu/iiif/auth/v2/probe?id=https://stacks.stanford.edu/file/bh502xm3351/CAS_bh502xm3351.pdf

This fixes it by allowing our parsing of incoming URIs to work without a druid: prefix.

### DIFF
--- a/spec/requests/iiif/auth/v2/probe_service_spec.rb
+++ b/spec/requests/iiif/auth/v2/probe_service_spec.rb
@@ -3,6 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe 'IIIF auth v2 probe service' do
+  let(:probe_service) { Iiif::Auth::V2::ProbeServiceController }
   let(:id) { 'bb461xx1037' }
   let(:file_name) { 'SC0193_1982-013_b06_f01_1981-09-29.pdf' }
   let(:stacks_uri) { "https://stacks-uat.stanford.edu/file/druid:#{id}/#{URI.encode_uri_component(file_name)}" }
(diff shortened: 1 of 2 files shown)


spec/requests/iiif/auth/v2/probe_service_spec.rb
```diff
@@ -91,6 +92,19 @@
...
+
+        it 'returns a success response' do
+          expect(response).to have_http_status :ok
+          expect(response.parsed_body).to include({
+                                                    "@context" => "http://iiif.io/api/auth/2/context.json",
+                                                    "type" => "AuthProbeResult2",
+                                                    "status" => 200
+                                                  })
+        end
+      end
```
peetucket, line 107: note that due to our mocking, this test is not quite comprehensive enough, hence the testing below of the actual parsing method

spec/requests/iiif/auth/v2/probe_service_spec.rb
```diff
@@ -388,5 +402,21 @@
                                               })
     end
   end
+
+  describe '#parse_uri' do
```
jcoyne, line 406: This seems like abusing request tests.  If we need to test this, can you extract this to a service and test it separately?

jcoyne (changes requested):

spec/requests/iiif/auth/v2/probe_service_spec.rb
jcoyne: How about something like https://github.com/sul-dlss/stacks/pull/1090 as an alternative way of testing this fix.

peetucket:
Closed in favor of #1090

PR #1090, opened 2023-12-15
Ensure that StacksFile is initialized without a druid prefix

jcoyne:


peetucket (approved):
Looks good.  I'll close #1089

PR #1091, opened 2023-12-15
Add a validator on stacks files

jcoyne:
This will give us a better message if the druid: prefix should sneak in

### DIFF
--- a/app/controllers/iiif/auth/v2/probe_service_controller.rb
+++ b/app/controllers/iiif/auth/v2/probe_service_controller.rb
@@ -7,7 +7,7 @@ module Iiif
       # Check access for IIIF auth v2
       # https://iiif.io/api/auth/2.0/#probe-service
       class ProbeServiceController < ApplicationController
-        def show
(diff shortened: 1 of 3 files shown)


peetucket:
If we are not expecting stack URLs to get files to have the "druid:" prefix in them, shouldn't we remove it from sul-embed here?  https://github.com/sul-dlss/sul-embed/blob/main/app/models/embed/purl/resource_file.rb#L40

Right now, all of the links generated in PURL pages to download files have that druid: prefix added.

jcoyne:
@peetucket that doesn't matter, because we remove the prefix in the controller before initializing the `StacksFile`

peetucket:
> @peetucket that doesn't matter, because we remove the prefix in the controller before initializing the `StacksFile`

Ah, right, yes, I remembered that, but originally I thought this code would be triggered, but now I see the controller would remove it first.  So this really should not never happen via the API anyway.

peetucket (approved):

PR #1105, opened 2024-01-04
Upgrade to Rails 7.1

jcoyne:


config/environments/production.rb
```diff
@@ -34,24 +33,34 @@
   # config.asset_host = "http://assets.example.com"
 
   # Specifies the header that your server uses for sending files.
-  config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
+  # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
```
corylown, line 36: I think we want to set this header: https://github.com/sul-dlss/stacks/pull/898
```suggestion
  config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
```

corylown (changes requested):

config/environments/production.rb
jcoyne: Thanks! Got it fixed.

corylown (approved):

PR #1108, opened 2024-01-16
Add missing gems to Gemfile.lock

corylown:


jcoyne (approved):

peetucket (approved):
I wonder why these gems were not added as dependencies in gemfile.lock if they are required by the rails 7 upgrade

jcoyne:
I'm also curious as to what is going on.

jcoyne:
It looks like dependency updates removed it: https://github.com/sul-dlss/stacks/pull/1107/files#diff-89cade48462044ee1b672dc5f4c3ec250fbd29effcd8932096a23c1283c6731fL74

peetucket:
I thought that if a gem (e.g. something in rails 7) needed them again, then the bundle install would have added them into gemfile.lock as dependencies even if they weren't explicitly added in the Gemfile itself.

PR #1147, opened 2024-05-16
Read files from OCFL root when feature flag is enabled

### LINKED ISSUE #1140: Read files from the head version of the OCFL directory
Add a feature flag that would enable lookup of files and metadata in the OCFL directory.  Ensure this is off by default.  Create a separate testing config with this on and one for it off.  For now assume that we only need this when it's reading metadata from JSON.

mjgiarlo:
Read files from OCFL root when feature flag is enabled

Fixes #1140

These commits makes it possible for Stacks to read files from an OCFL-structured storage root, and puts that logic behind a new feature flag. To accomplish this, it encapsulates knowledge of OCFL vs. legacy stacks structures within a new `StorageRoot` class.


app/models/stacks_file.rb
```diff
@@ -29,21 +29,15 @@ def content_length
   end
 
   def path
-    @path ||= begin
-      return unless treeified_path
-
-      File.join(Settings.stacks.storage_root, treeified_path)
-    end
```
mjgiarlo, line 36: Move storage root logic into a dedicated class.

app/models/stacks_file.rb
```diff
@@ -29,21 +29,15 @@ def content_length
...
+    @path ||= storage_root.absolute_path
   end
 
   def treeified_path
-    return unless druid_parts && file_name
-
-    File.join(druid_parts[1..4], file_name)
+    storage_root.relative_path
   end
```
mjgiarlo, line 37: Move treeification concern into new `StorageRoot` class

spec/rails_helper.rb
```diff
@@ -46,6 +46,12 @@
   # instead of true.
   config.use_transactional_fixtures = true
 
+  config.before do
+    allow_any_instance_of(StorageRoot) # rubocop:disable RSpec/AnyInstance
+      .to receive(:version_directories)
```
jcoyne, line 51: It doesn't feel good to have this kind of global stub.  Why do we need it?

jcoyne (changes requested):

spec/rails_helper.rb
mjgiarlo: I wrestled with this a bunch and was unsatisfied with this solution myself. Here's what's up:

The logic in `StorageRoot` for OCFL structures requires there actually being directories on the filesystem, because with OCFL we need the head version in the path which means we need to *find* the head version (unless the logic for the legacy structure which is happy to emit paths that do not exist). Maybe adding the `ocfl-rb` gem will give us a better way to handle this or stub it out in RSpec-land?

app/models/storage_root.rb
```diff
@@ -0,0 +1,61 @@
...
+  def versioning_enabled?
+    Settings.features.read_stacks_from_ocfl_root
+  end
+
+  def path_parts
+    [
+      treeified_id,
+      file_name
+    ].tap do |parts|
+      parts.insert(1, head_version, 'content') if versioning_enabled?
```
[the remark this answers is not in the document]
jcoyne: Yes, but we may need to add that to the gem first. I'm not sure it supports that yet.

mjgiarlo:
@jcoyne All changes made. Thanks!

jcoyne (approved):

PR #1149, opened 2024-05-17
Remove CDL auth

jcoyne:
This hasn't been used in a few years. It will be hard to work in versioning and maintain CDL at the same time without understanding what the modern CDL requirements are.


justinlittman:
Has @andrewjbtw blessed?

justinlittman (approved):

PR #1160, opened 2024-05-23
Bump OCFL to 0.6.0 so we can retrieve files properly

### LINKED ISSUE #1159: Bump OCFL gem and adjust code to use new API for getting the latest version of a file
~~Blocked by https://github.com/sul-dlss/ocfl-rb/issues/17~~

mjgiarlo:
Fixes #1159


app/models/storage_root.rb
```diff
@@ -65,7 +65,7 @@ def relative_path
     end
 
     def absolute_path
-      object_root.path(:head, @file_name)
+      object_root.path(filepath: @file_name)
```
jcoyne, line 68: Shouldn't we be specifying that it is present in the `version: :head`?  I don't think we want to serve it if it was deleted in a later version.

mjgiarlo: @jcoyne If you call `Directory#path` without the `version` kwarg, the `ocfl` gem will look up a file path in the object inventory manifest. If a file has been deleted, it should not be in the object inventory manifest. It will only be referenced in version state in the manifest.

mjgiarlo: (This was the point of doing https://github.com/sul-dlss/ocfl-rb/pull/27)

mjgiarlo: @jcoyne Does that make sense or do you think we've gotten something wrong?

mjgiarlo: Made the change. This should be good now.

jcoyne (approved):

PR #1189, opened 2024-06-28
[refactor] Pass StacksFile into the StacksImage

jcoyne:
This means that StacksImage doesn't need to know how to create its own StacksFile.

This also makes the initializer of StacksImage explicit, which helps the reader to understand which attributes are required.

This supports moving to content addressable storage.


cbeer:
If we're going to do this, can we also refactor the other models (StacksMediaStream .... unclear to me if StacksMediaToken needs it) to match?

jcoyne:
@cbeer I did StacksMediaStream, but I don't think we want to extend this into StacksMediaToken because then `create_from_encrypted_string` would need to be aware of StacksFile.

PR #1219, opened 2024-07-29
Use version when requesting cocina from purl endpoint

### LINKED ISSUE #1213: Retrieve metadata from PURLs versioned endpoint
Assume the version is 1 for now.

Depends on:
https://github.com/sul-dlss/purl/issues/1077

### LINKED ISSUE #1214: Add a file route that allows requestor to specify a version
Since the `file_name` can be hierarchical, make a totally different base route to avoid any possible name collisions.

I recommend something like:
```
  constraints id: druid_regex do
    namespace 'v2' do
      get '/file/:id/:version_id/*file_name' => 'versions#show', format: false
    end
  end
```

The `V2::VersionsController` can be a copy of the `FilesController`, but it will lookup the metadata and file from the versioned path.   If the `version_id` is "1", and no versioned path is found, it should fall back to the legacy path for the file.

aaron-collier:
Closes #1213 
Closes #1214


jcoyne (changes requested):
What about this route? https://github.com/sul-dlss/stacks/blob/2a5595fef35f0938d1d43e1a16e9fac054ebe135/app/controllers/file_controller.rb#L77 This is needed for a single file view.

app/controllers/object_controller.rb
```diff
@@ -38,6 +38,10 @@ def druid
     params[:id]
   end
 
+  def version
+    params[:version] || 'v1'
```
cbeer, line 42: Is `v1` the right default? Shouldn't this be "whatever the latest head version is"?

app/controllers/v2/versions_controller.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+
+module V2
+  # API for delivering files from stacks by version
+  class VersionsController < ApplicationController
```
cbeer, line 5: I guess I don't understand why we need a new controller (that's nearly the same as `FileController`). That seems unfortunate.

aaron-collier: That was the suggestion in the ticket. Happy to revert (maybe can be a standup discussion or a follow up pr if this is blocking)

jcoyne: I thought this was the easiest way to disambiguate the route. The current route allows slashes in filename, so adding a version there may have the potential of collision with a directory or filename.

cbeer: A separate route seems fine, I just don't know why we need a nearly identical controller to go with it.

jcoyne: I can go either way on that.  One controller is fine by me.

aaron-collier: @jcoyne / @cbeer - I removed the second controller.

app/controllers/file_controller.rb
```diff
@@ -74,6 +74,10 @@ def current_file
   end
 
   def cocina
-    @cocina ||= Cocina.find(params[:id])
+    @cocina ||= Cocina.find(params[:id], version)
+  end
+
+  def version
+    params[:version_id] || :head
```
cbeer, line 81: Should `version_id` also be permitted in `file_params` (so the bounce through auth can work)?

spec/requests/versioned_file_spec.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe "Versioned File requests" do
+  before do
+    allow(Cocina).to receive(:find).and_return(Cocina.new(public_json))
```
cbeer, line 7: Should the test also make sure we're retrieving the "right" version of the cocina?

app/controllers/file_controller.rb
aaron-collier: Hmm, I'm not sure about `WebauthController` as it symbolizes the keys I think it should be good as is, but open to suggestion.

spec/requests/versioned_file_spec.rb
aaron-collier: Done.

app/controllers/file_controller.rb
cbeer: Don't we need to check if there's a version parameter and send it to the right route? Right now, it looks like it will always redirect to the non-versioned path?

https://github.com/sul-dlss/stacks/blob/main/app/controllers/webauth_controller.rb#L32

aaron-collier: @cbeer ah I see, I thought that the params would determine that, but walking through a debugger I see it passes it as a query param not part of the path. Fixed.

justinlittman (approved):

PR #1225, opened 2024-08-05
Use content addressed storage file when available.

jcoyne:


app/models/storage_root.rb
```diff
@@ -38,26 +26,41 @@ def treeified_id
...
   def path_finder_class
     LegacyPathFinder
   end
 
+  def druid_parts
+    @druid_parts ||= druid.match(DRUID_PARTS_PATTERN)
+  end
+
   # Calculate file paths in the legacy Stacks structure
   class LegacyPathFinder
```
justinlittman, line 41: Why is this a LegacyPathFinder?

justinlittman (approved):
Approve pending consideration of comment.

PR #1316, opened 2024-11-26
Remove dependency on zipline

### LINKED ISSUE #1311: zipline doesn't seem to be maintained
It doesn't work on Rails 8:
https://github.com/fringd/zipline

jcoyne:
Because zipline is not yet compatable with the latest version of Rails.  It also has many features that we don't use. Fixes #1311

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,7 +73,7 @@ gem 'http'
 gem 'cancancan'
 gem 'dalli'
 gem 'retries'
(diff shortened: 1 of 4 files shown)


andrewjbtw:
Will this break our zip download feature?

jcoyne:
@andrewjbtw no

thatbudakguy (approved):

PR #1320, opened 2024-12-03
Upgrade to Rails 8

jcoyne:


dnoneill (changes requested):
remove icon files

jcoyne:
This is only updating existing icon files, which already exist.  The removal is happening in https://github.com/sul-dlss/stacks/pull/1359 as they are no longer used when we switch to the icons in the component library.

jcoyne:
@dnoneill Icon files have been removed

dnoneill (approved):

PR #1321, opened 2024-12-03
Use builtin bundler caching for the build.

jcoyne:
This is a blocker for #1320

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -18,15 +18,17 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
-      - uses: actions/cache@v4
-        with:
-          path: vendor/bundle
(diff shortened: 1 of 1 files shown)


.github/workflows/ruby.yml
```diff
@@ -18,15 +18,17 @@ jobs:
...
-        run: bin/rake
+        run: bin/rake test:prepare spec
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.3
```
thatbudakguy, line 31: should this use `${{ matrix.ruby }}` too?

jcoyne: No, we only want to run the linter one time.

thatbudakguy: I guess there's theoretically a case where those could differ if we added other ruby versions to test.

thatbudakguy (approved):

PR #1400, opened 2025-03-10
add wms location for geo login

dnoneill:
part of https://github.com/sul-dlss/sul-embed/issues/2234

### DIFF
--- a/app/controllers/iiif/auth/v2/probe_service_controller.rb
+++ b/app/controllers/iiif/auth/v2/probe_service_controller.rb
@@ -7,7 +7,7 @@ module Iiif
       # Check access for IIIF auth v2
       # https://iiif.io/api/auth/2.0/#probe-service
       class ProbeServiceController < ApplicationController
-        def show # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+        def show # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/PerceivedComplexity
           # Example call:
           # /iiif/auth/v2/probe?id=https://stacks-uat.stanford.edu/file/druid:bb461xx1037/folder/SC0193_1982-013_b06_f01_1981-09-29.pdf
           stacks_uri = params[:id] # this is a fully qualified URI to the resource on the stacks that the user is requesting access to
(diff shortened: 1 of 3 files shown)


app/controllers/iiif/auth/v2/probe_service_controller.rb
```diff
@@ -26,15 +27,11 @@ def show # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
           elsif !file.readable?
             json[:status] = 404
           elsif can? :access, file
-            if file.streamable?
+            is_geo = cocina.data['type'] == 'https://cocina.sul.stanford.edu/models/geo'
```
jcoyne, line 30: Please move this to `Cocina#geo?`

app/controllers/iiif/auth/v2/probe_service_controller.rb
```diff
@@ -46,6 +43,18 @@ def show # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
           render json:
         end
 
+        def iiif_location(is_geo, file)
+          # 4 hour token
+          token = JWT.encode({ data: 'geo_token', exp: Time.now.to_i + (4 * 3600) }, Settings.geo.proxy_secret, 'HS256')
```
jcoyne, line 48: This line is happening even for media objects.  Can this be in the `is_geo` conditional?

jcoyne (changes requested):

app/controllers/iiif/auth/v2/probe_service_controller.rb
dnoneill: Done

app/controllers/iiif/auth/v2/probe_service_controller.rb
dnoneill: I ended up refactoring this a bit to do this but also not have separate return statements since we no longer need them because video and geo both use a token now

app/controllers/iiif/auth/v2/probe_service_controller.rb
```diff
@@ -46,6 +42,20 @@ def show # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
           render json:
         end
 
+        def iiif_location(is_geo, file)
+          if is_geo
+            # 4 hour token
+            token = JWT.encode({ data: 'geo_token', exp: Time.now.to_i + (4 * 3600) }, Settings.geo.proxy_secret, 'HS256')
+            url = Settings.geo.proxy_url
```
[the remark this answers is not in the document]
dnoneill: No. sul-embed does the identifier. https://github.com/sul-dlss/sul-embed/blob/61cc8596e9eb8832b4b3f34d86a70ea0d160d36f/app/javascript/src/modules/geo_viewer.js#L71, layers = the druid

jcoyne (approved):

PR #1405, opened 2025-03-17
return 403 for no download on pdf, 3d

dnoneill:
part of https://github.com/sul-dlss/sul-embed/issues/2783

### DIFF
--- a/app/controllers/iiif/auth/v2/probe_service_controller.rb
+++ b/app/controllers/iiif/auth/v2/probe_service_controller.rb
@@ -48,6 +48,7 @@ module Iiif
 
         private
 
+        # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
         def auth_probe_result(file, cocina)
           if !file.valid?
             AuthProbeResult2.bad_request(file.errors.full_messages)
@@ -59,21 +60,33 @@ module Iiif
             else
               AuthProbeResult2.ok
             end
+          elsif file.stanford_restricted?
+            AuthProbeResult2.unauthorized(**unauthorized_fields(file))
           else
-            AuthProbeResult2.unauthorized(unauthorized_heading(file))
+            AuthProbeResult2.forbidden(**forbidden_fields(file))
+          end
+        end
+        # rubocop:enable Metrics/AbcSize, Metrics/PerceivedComplexity
+
+        def forbidden_fields(file)
+          if file.embargoed?
+            { heading: I18n.t('probe_service.embargoed', date: file.embargo_release_date.to_date),
+              icon: I18n.t('probe_service.embargoed_icon') }
+          elsif file.restricted_by_location?
+            { heading: I18n.t('probe_service.location', location: Settings.user.locations.labels.send(file.location)),
+              icon: I18n.t('probe_service.location_icon') }
+          else
+            { heading: I18n.t("probe_service.no_download"), icon: I18n.t('probe_service.no_download_icon') }
           end
         end
 
-        # add details to response for when access is denied
-        def unauthorized_heading(file)
-          if file.stanford_restricted? && !file.embargoed?
-            I18n.t('probe_service.stanford')
-          elsif file.stanford_restricted? && file.embargoed?
-            I18n.t('probe_service.stanford_and_embargoed', date: file.embargo_release_date.to_date)
-          elsif file.embargoed?
-            I18n.t('probe_service.embargoed', date: file.embargo_release_date.to_date)
-          elsif file.restricted_by_location?
-            I18n.t('probe_service.location', location: Settings.user.locations.labels.send(file.location))
+        # add details to response for when stanford_restricted?
+        def unauthorized_fields(file)
+          if file.embargoed?
+            { heading: I18n.t('probe_service.stanford_and_embargoed', date: file.embargo_release_date.to_date),
+              icon: I18n.t('probe_service.embargoed_icon') }
+          else
+            { heading: I18n.t('probe_service.stanford'), icon: I18n.t('probe_service.stanford_icon') }
           end
         end
 
--- a/app/models/auth_probe_result2.rb
+++ b/app/models/auth_probe_result2.rb
@@ -32,15 +32,16 @@ class AuthProbeResult2
 
   # When not authorized to view the resource.
   class ErrorResult < AuthProbeResult2
-    def initialize(heading:, **)
+    def initialize(heading:, icon: nil, **)
       super(status: 401, **)
       @heading = heading
+      @icon = icon
     end
 
-    attr_reader :heading
+    attr_reader :heading, :icon
 
     def as_json
-      super.merge(heading:)
+      super.merge(heading:, icon:)
     end
   end
 
@@ -52,8 +53,16 @@ class AuthProbeResult2
     new(status: 400, note: { en: messages })
   end
 
-  def self.unauthorized(heading)
-    ErrorResult.new(note: { en: [I18n.t('probe_service.access_restricted')] }, heading: { en: [heading] })
+  # return 403 since there is nothing a user can do to get access
+  # IIIF has the API return 401 for views that can be logged into,
+  # which we don't want for no_download, embargo with no stanford login, location restricted
+  # https://iiif.io/api/auth/2.0/#71-authorization-flow-algorithm
+  def self.forbidden(heading:, icon:)
+    ErrorResult.new(status: 403, note: { en: [I18n.t('probe_service.access_restricted')] }, heading: { en: [heading] }, icon:)
+  end
+
+  def self.unauthorized(heading:, icon:)
+    ErrorResult.new(note: { en: [I18n.t('probe_service.access_restricted')] }, heading: { en: [heading] }, icon:)
   end
 
   def self.not_found(druid)
--- a/app/models/cocina.rb
+++ b/app/models/cocina.rb
@@ -57,6 +57,13 @@ class Cocina
     type == 'https://cocina.sul.stanford.edu/models/geo'
   end
 
+  # types that cannot be viewed when download = 'none'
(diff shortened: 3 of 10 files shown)


app/models/stacks_file.rb
```diff
@@ -55,7 +55,18 @@ def stacks_rights
   end
 
   delegate :rights, :restricted_by_location?, :stanford_restricted?, :embargoed?,
-           :embargo_release_date, :location, to: :stacks_rights
+           :embargo_release_date, :location, :no_download?, :world_viewable?, to: :stacks_rights
+
+  # types that cannot be viewed when download = 'none'
+  def check_download_types
+    ['https://cocina.sul.stanford.edu/models/document', 'https://cocina.sul.stanford.edu/models/3d']
```
jcoyne, line 62: Can we put these types in the cocina.rb, similar to how we already have for geo? https://github.com/sul-dlss/stacks/blob/main/app/models/cocina.rb#L56C1-L58

jcoyne (changes requested):

app/models/stacks_file.rb
jcoyne: We may want to consider making a method like `Cocina#proxied?`, which would return true for media or geo.  Stacks would evaluate the download permissions if they are not proxied.

dnoneill: The reason I went with this pdf/3d is because I am unsure about images and those are three types. I was assuming they were what you are calling proxied.

jcoyne: I'm calling things proxied if they have a location in the probe service response: https://iiif.io/api/auth/2.0/#probe-service-response

dnoneill: Gotcha. Yeah I think this is a good idea assuming I can get clarity on how images work when they are set to download=none. I think they still are viewable.

app/controllers/iiif/auth/v2/probe_service_controller.rb
```diff
@@ -80,12 +80,21 @@ def add_detail(file)
...
+            detail[:status] = 403
+            detail[:heading] = { en: [heading_html('no_download')] }
           end
           detail[:note] = { en: [I18n.t('probe_service.access_restricted')] }
           detail
         end
-        # rubocop:enable Metrics/AbcSize
+
+        def heading_html(heading_field, location: nil)
+          "#{I18n.t("probe_service.#{heading_field}_icon")}<p class='loginMessage'>#{I18n.t("probe_service.#{heading_field}",
```
jcoyne, line 94: Can we leave the markup out of this PR?  This is going to couple the CSS across two different projects.  In my experience that gets' hard to follow, especially when we're not actively paying attention to both codebases, or a developer who is less familiar with this gets assigned to work on this.

jcoyne (changes requested):

app/controllers/iiif/auth/v2/probe_service_controller.rb
dnoneill: Done. Moved svg into icons. This makes a much cleaner code all around

app/models/stacks_file.rb
dnoneill: made the field not_proxied? works for this

app/models/stacks_rights.rb
```diff
@@ -22,6 +22,20 @@ def stanford_restricted?
...
 
+  def no_download?
+    rights.download == 'none'
+  end
+
+  def world_viewable?
+    rights.view == 'world'
+  end
+
+  def viewable?
```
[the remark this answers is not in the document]
dnoneill: You are suggesting in the cocina_ability updating it to:

    can [:access], access_models do |f|
      return f.world_viewable? unless f.cocina.not_proxied?

      f.world_viewable? && !f.no_download?
    end

app/models/cocina.rb
```diff
@@ -57,6 +57,11 @@ def geo?
     type == 'https://cocina.sul.stanford.edu/models/geo'
   end
 
+  # types that cannot be viewed when download = 'none'
+  def not_proxied?
```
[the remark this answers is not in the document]
dnoneill: I don't know. I think no? I am not quite sure. We don't use the probe service for files. :access is only used for the probe service and media_authenication which hopefully we can get rid of soon. We can add it later but I am not actually sure how the download works for files.

app/models/stacks_rights.rb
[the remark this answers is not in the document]
dnoneill: I am assuming that is what you meant. Updated.

app/controllers/iiif/auth/v2/probe_service_controller.rb
```diff
@@ -59,8 +60,22 @@ def auth_probe_result(file, cocina)
...
+        end
+        # rubocop:enable Metrics/AbcSize, Metrics/PerceivedComplexity
+
+        def forbidden_fields(file)
+          if file.embargoed?
+            [I18n.t('probe_service.embargoed', date: file.embargo_release_date.to_date), I18n.t('probe_service.embargoed_icon')]
+          elsif file.restricted_by_location?
+            [I18n.t('probe_service.location', location: Settings.user.locations.labels.send(file.location)),
+             I18n.t('probe_service.location_icon')]
+          elsif file.no_download?
```
jcoyne, line 77: Do we need a conditional here?  Isn't this the only possible outcome?

app/controllers/iiif/auth/v2/probe_service_controller.rb
```diff
@@ -59,8 +60,22 @@ def auth_probe_result(file, cocina)
...
-          else
+          elsif file.stanford_restricted?
             AuthProbeResult2.unauthorized(unauthorized_heading(file))
+          else
+            AuthProbeResult2.forbidden(*forbidden_fields(file))
+          end
+        end
+        # rubocop:enable Metrics/AbcSize, Metrics/PerceivedComplexity
+
+        def forbidden_fields(file)
```
jcoyne, line 71: Instead of a tuple, can we return a hash with `:note` and `:icon` keys?  I think the tuple makes it a bit more challenging to follow.

app/controllers/iiif/auth/v2/probe_service_controller.rb
```diff
@@ -70,10 +85,6 @@ def unauthorized_heading(file)
             I18n.t('probe_service.stanford')
           elsif file.stanford_restricted? && file.embargoed?
             I18n.t('probe_service.stanford_and_embargoed', date: file.embargo_release_date.to_date)
-          elsif file.embargoed?
```
jcoyne, line 73: Why drop embargoed?  Aren't there embargos where the file is stanford-only and you can login to see it.  When the embargo is lifted it will be world readable.

app/models/auth_probe_result2.rb
```diff
@@ -52,6 +54,14 @@ def self.bad_request(messages)
     new(status: 400, note: { en: messages })
   end
 
+  # return 403 since there is nothing a user can do to get access
+  # IIIF has the API return 401 for views that can be logged into,
+  # which we don't want for no_download, embargo with no stanford login, location restricted
+  # https://iiif.io/api/auth/2.0/#71-authorization-flow-algorithm
+  def self.forbidden(heading, icon)
```
jcoyne, line 61: Can we make these kwargs?

app/models/cocina_ability.rb
```diff
@@ -49,7 +49,7 @@ def initialize(user)
     end
 
     can [:access], access_models do |f|
-      f.rights.view == 'world'
+      f.world_viewable? && (f.not_proxied? ? !f.no_download? : true)
```
jcoyne, line 52:
```suggestion
      f.world_viewable? && (!f.not_proxied? || !f.no_download?)
```

jcoyne (changes requested):

app/controllers/iiif/auth/v2/probe_service_controller.rb
dnoneill: I was considering that but the original code didn't have an else on headings. I don't think we necessarily want to just do else for everything https://github.com/sul-dlss/stacks/blob/760c866ee1065a12ce7746121446029f3911803b/app/controllers/iiif/auth/v2/probe_service_controller.rb#L68-L78.

app/controllers/iiif/auth/v2/probe_service_controller.rb
dnoneill: The two lines before check for stanford_restricted and embargoed. We moved the embargoed into forbidden_fields because if it gets to embargoed then the file is embargoed and not stanford restricted which means no one can look at it.

app/controllers/iiif/auth/v2/probe_service_controller.rb
jcoyne: The other option is that it returns `nil` here, and we aren't testing or expecting that. I'd rather have deterministic behavior (even if we have to fix it later).

app/controllers/iiif/auth/v2/probe_service_controller.rb
jcoyne: Ah. I see now.

app/controllers/iiif/auth/v2/probe_service_controller.rb
dnoneill: updated

dnoneill:
@jcoyne Is this ready to go?

app/models/auth_probe_result2.rb
```diff
@@ -3,16 +3,18 @@
 # Represents IIIF Auth V2 probe service response
 # See https://iiif.io/api/auth/2.0/#probe-service-response
 class AuthProbeResult2
-  def initialize(status:, note: nil)
+  def initialize(status:, note: nil, icon: nil)
```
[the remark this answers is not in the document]
dnoneill: @jcoyne That makes sense. Done.

jcoyne (approved):

PR #1406, opened 2025-03-17
[refactor] Extract AuthProbeResult2 class

jcoyne:
This fixes the rubocop errors and separates the formatting of the response out of the controller. 

This also removes `auth_url`, which is not a valid property of the probe service response (https://iiif.io/api/auth/2.0/#probe-service-response) and was not used.


dnoneill:
I would double check that we aren't using auth_url for sul-embed. There is a variable that I am pretty sure is being set in sul-embed via a setting but I would double check we don't break all our media items.

jcoyne:
I have double checked.

dnoneill (approved):

PR #1408, opened 2025-03-18
Return a message for citation-only files from the iiif v2 auth probe …

jcoyne:
…service


dnoneill:
@jcoyne https://github.com/sul-dlss/stacks/pull/1405 will does this

PR #1423, opened 2025-03-28
fix degraded info.json to display correct dimensions, sizes

dnoneill:
part of #1422 

Before degraded info.json
Screenshot 2025-03-28 at 2 10 52 PM

After
Screenshot 2025-05-28 at 11 39 11 AM

### DIFF
--- a/spec/requests/iiif_spec.rb
+++ b/spec/requests/iiif_spec.rb
@@ -6,17 +6,15 @@ RSpec.describe 'IIIF API' do
   let(:metadata) do
     { 'height' => 2099,
       'width' => 1702,
+      'sizes' => [{ 'width' => 1702, 'height' => 2099 }],
       'tiles' => [{ 'width' => 256, "height" => 256, "scaleFactors" => [1, 2, 4, 8, 16] }] }
-  end
-  let(:metadata_service) do
-    instance_double(IiifMetadataService, fetch: metadata,
-                                         image_width: 1702,
-                                         image_height: 2552)
   end
 
   before do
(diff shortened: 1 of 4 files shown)


jcoyne (changes requested):
Shoudn't the `@id` property also change to the degraded one?  Currently it's referring to the original image, so it shouldn't show degraded sizes.

jcoyne (changes requested):
Can you please add a test that shows the change to `IIIFController#canonical_url`?

dnoneill:
@jcoyne So it turns out that  spec/requests/iiif_spec.rb isn't actually working as I would expect. When I run 

```
 it 'returns degraded info.json' do
   get '/image/iiif/degraded/nr349ct7889/image/info.json'
   expect(response).to have_http_status :ok
   puts JSON.parse(response.body).keys.inspect
end
          ```
 There is no id. I think there is some mocking being done which means this isn't the right place to test the id function.
 
I tried in the iiif_info_service_spec.rb and set degraded to false but I ran into an issue where it was telling me to stub the response which I did and it didn't seem to want to catch it.

File: spec/requests/iiif_spec.rb [hunk 1]
```diff
@@ -141,7 +140,7 @@
           get '/image/iiif/bb000cr7262%2Fimage/info.json'
           json = response.parsed_body
 
-          expect(json['sizes']).to eq [{ 'width' => 266, 'height' => 400 }]
+          expect(json['sizes']).to eq [{ 'width' => 324, 'height' => 400 }]
```
dnoneill, line 143: This is actually the correct dimension to preserve aspect ratio

PR #1426, opened 2025-03-31
move note without heading to heading field

### LINKED ISSUE #1425: Probe service note field
According to the IIIF probe spec, the note field can't appear without the heading field. https://iiif.io/api/auth/2.0/#note-1. We need to migrate our note field to a heading field.

dnoneill:
closes #1425

### DIFF
--- a/app/models/auth_probe_result2.rb
+++ b/app/models/auth_probe_result2.rb
@@ -3,15 +3,17 @@
 # Represents IIIF Auth V2 probe service response
 # See https://iiif.io/api/auth/2.0/#probe-service-response
 class AuthProbeResult2
-  def initialize(status:, note: nil)
+  def initialize(status:, note: nil, heading: nil)
     @status = status
     @note = note
+    @heading = heading
   end
 
-  attr_reader :status, :note
+  attr_reader :status, :note, :heading
 
   def as_json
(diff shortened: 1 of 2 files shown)


app/models/auth_probe_result2.rb
```diff
@@ -3,15 +3,17 @@
 # Represents IIIF Auth V2 probe service response
 # See https://iiif.io/api/auth/2.0/#probe-service-response
 class AuthProbeResult2
-  def initialize(status:, note: nil)
+  def initialize(status:, note: nil, heading: nil)
```
jcoyne, line 6: We only use note an heading on error results, so can these fields move/remain on `ErrorResult`?

jcoyne (changes requested):

app/models/auth_probe_result2.rb
dnoneill: And then not_found and bad_request become ErrorResults instead of AuthProbeResult2

jcoyne: Ah, I thought they were already.  Nevermind then. `ErrorResult` was perhaps too general of a name and it ought to have been `NotAuthorizedResponse`

jcoyne (approved):

PR #1456, opened 2025-07-31
Remove the X-Frame-Options header from the probe service

jcoyne:
This is a bit of a shot in the dark to fix https://github.com/sul-dlss/sul-embed/issues/2894


jcoyne:
This did not fix the problem.

PR #1481, opened 2025-08-07
In ApprovedLocation, locatable is always a user

jcoyne:


app/models/approved_location.rb
```diff
@@ -5,21 +5,19 @@
 # based on a provided locatable object's IP address
```
dnoneill, line 5: This comment should be updated

PR #1505, opened 2025-09-04
encode inline text files for the browser

dnoneill:


app/controllers/file_controller.rb
```diff
@@ -24,10 +24,21 @@ def show
...
+    send_file current_file.path, filename: current_file.file_name, type: file_type, disposition:
   end
   # rubocop:enable Metrics/AbcSize
 
+  def file_type
+    mimetype = current_file.cocina.find_file(current_file.file_name)['hasMimeType']
+    return mimetype unless mimetype.starts_with?('text')
+
+    content = File.read(current_file.path)
+    encoding = content.encoding.name
```
jcoyne, line 36: Encoding is a property of the environment, not the file.

for example:
```
➜ ruby -e 'p File.read("Gemfile").encoding'
#<Encoding:UTF-8>
➜  ruby -E ISO-8859-1 -e 'p File.read("Gemfile").encoding'
#<Encoding:ISO-8859-1>
```

PR #1524, opened 2025-11-04
Cache the json response for 5 minutes

taylor-steve:
When a user visits a purl page, currently all initial requests are going through a single stacks node. This might not be intentional or dependable behavior (e.g., it could be a byproduct of specific load balancer configuration). 

However, for this case, caching the json response locally saves `n` info.json requests to the image servers, where `n` is the number of tiles being requested. Even in the worst case we are saving 1 request.

### DIFF
--- a/app/services/iiif_metadata_service.rb
+++ b/app/services/iiif_metadata_service.rb
@@ -66,7 +66,7 @@ class IiifMetadataService
   end
 
   def json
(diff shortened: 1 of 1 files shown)


jcoyne:
Okay, I was confused by the inline code comments that rails generated. I think you are right.

PR #1579, opened 2026-04-21
Stop limiting IIIF requests to 256px tiles

jcoyne:


taylor-steve:
Noting for our future-selves. Vivian to me she talked to Tom and Tom says the tile restriction needs to stay in place, for now.

jcoyne:
Would it be permitted for the user to request 4 adjacent 256px tiles in a single request?

PR #1583, opened 2026-05-14
set file content type in header for head request

### LINKED ISSUE #1569: Responses to HTTP HEAD requests should return the content type of the file that would be GET-ed
An SDR depositor reported that whenever they make a HEAD request for a javascript file they deposited, the server responds with a content type of `text/html`:

```
$ curl -I https://stacks.stanford.edu/file/wy642jd6829/quarto-live_files/libs/quarto-contrib/live-runtime/pyodide-worker.js
HTTP/1.1 200 OK
Date: Tue, 06 Jan 2026 01:12:35 GMT
Server: Apache/2.4.52 (Ubuntu)
access-control-allow-origin: *
cache-control: max-age=600, private
strict-transport-security: max-age=31536000; includeSubDomains
x-request-id: 26ae53b1-2579-43f4-83c4-4b987df33d79
accept-ranges: bytes
x-runtime: 0.029346
Last-Modified: Mon, 29 Dec 2025 17:00:51 GMT
Transfer-Encoding: chunked
etag: W/"6f8c0b38e328510837f9c2b15d3d6386"
Status: 200 OK
Content-Type: text/html
```

After trying a few different kinds of requests, I think the problem is that Stacks responses to HEAD requests always return `text/html` regardless of the requested file's type. When you issue a GET request, Stacks responds with the mimetype of the resource being requested (which I think matches the cocina metadata).

```
$ curl -v https://stacks.stanford.edu/file/wy642jd6829/quarto-live_files/libs/quarto-contrib/live-runtime/pyodide-worker.js 1>/dev/null
> GET /file/wy642jd6829/quarto-live_files/libs/quarto-contrib/live-runtime/pyodide-worker.js HTTP/1.1
> Host: stacks.stanford.edu
> User-Agent: curl/8.18.0
> Accept: */*
> 
* Request completely sent off
{ [5 bytes data]
< HTTP/1.1 200 OK
< Date: Mon, 23 Feb 2026 18:27:00 GMT
< Server: Apache/2.4.52 (Ubuntu)
< access-control-allow-origin: *
< cache-control: max-age=600, private
< access-control-expose-headers: Content-Range
< strict-transport-security: max-age=31536000; includeSubDomains
< x-request-id: 92e4ea86-e8fe-47af-a452-6d890a7ddc4d
< content-disposition: inline; filename="quarto-live_files%2Flibs%2Fquarto-contrib%2Flive-runtime%2Fpyodide-worker.js"; filename*=UTF-8''quarto-live_files%2Flibs%2Fquarto-contrib%2Flive-runtime%2Fpyodide-worker.js
< accept-ranges: bytes
< x-runtime: 0.029971
< Last-Modified: Mon, 29 Dec 2025 17:00:51 GMT
< etag: W/"6f8c0b38e328510837f9c2b15d3d6386"
< Status: 200 OK
< Vary: Accept-Encoding
< Transfer-Encoding: chunked
< Content-Type: application/x-javascript
```
You can see the same issue with HEAD requests for other file types, not just javascript:
```
$ curl -I https://stacks.stanford.edu/file/zv296hf8888/data.zip?download=true
HTTP/1.1 200 OK
Date: Mon, 23 Feb 2026 18:29:36 GMT
Server: Apache/2.4.52 (Ubuntu)
access-control-allow-origin: *
cache-control: max-age=600, private
access-control-expose-headers: Content-Range
strict-transport-security: max-age=31536000; includeSubDomains
x-request-id: 5fbdd8da-adde-48cd-8f82-8a9024f16bc1
accept-ranges: bytes
x-runtime: 0.117449
Last-Modified: Wed, 22 Nov 2017 01:35:27 GMT
Transfer-Encoding: chunked
etag: W/"50a4c4ce21161e8fb03107328878deaa"
Status: 200 OK
Content-Type: text/html
```

My understanding is that the response to an HTTP HEAD request should show the headers that would be sent if a GET was made for the same resource ([source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD)). So Stacks should be sending the content type of the file being requested. There may be other response header adjustments to make as well.

peetucket:
Fixes #1569 I think

Would be best to verify on stage to be sure

### DIFF
--- a/app/controllers/file_controller.rb
+++ b/app/controllers/file_controller.rb
@@ -63,8 +63,8 @@ class FileController < ApplicationController
     response.headers['Content-Range'] = "bytes #{range}/#{current_file.content_length}"
     response.headers['Content-Length'] = range.content_length.to_s
     if request.head?
-      set_file_headers
-      return head(:partial_content) # 206 Partial Content
+      set_head_response_headers
+      return head(:ok)
     end
 
     response.status = 206
@@ -83,7 +83,7 @@ class FileController < ApplicationController
   def handle_full_request
     response.headers['Content-Length'] = current_file.content_length.to_s
     if request.head?
-      set_file_headers
+      set_head_response_headers
       return head(:ok)
     end
 
@@ -98,7 +98,7 @@ class FileController < ApplicationController
     end
   end
 
-  def set_file_headers
+  def set_head_response_headers
     response.headers['Content-Type'] = current_file.content_type
     response.headers['Content-Disposition'] =
       ActionDispatch::Http::ContentDisposition.format(disposition: disposition, filename: current_file.file_name)
--- a/app/controllers/file_controller.rb
+++ b/app/controllers/file_controller.rb
@@ -2,6 +2,7 @@
 
 ##
 # API for delivering files from stacks
+# rubocop:disable Metrics/ClassLength
 class FileController < ApplicationController
   include ActionController::Live
 
@@ -61,7 +62,10 @@ class FileController < ApplicationController
 
     response.headers['Content-Range'] = "bytes #{range}/#{current_file.content_length}"
     response.headers['Content-Length'] = range.content_length.to_s
-    return head(:ok) if request.head?
+    if request.head?
+      set_file_headers
+      return head(:partial_content) # 206 Partial Content
+    end
 
     response.status = 206
 
@@ -78,7 +82,10 @@ class FileController < ApplicationController
 
   def handle_full_request
     response.headers['Content-Length'] = current_file.content_length.to_s
-    return head(:ok) if request.head?
+    if request.head?
+      set_file_headers
+      return head(:ok)
+    end
 
     send_stream(
       filename: current_file.file_name,
@@ -89,6 +96,12 @@ class FileController < ApplicationController
         stream.write(chunk)
       end
     end
+  end
+
+  def set_file_headers
+    response.headers['Content-Type'] = current_file.content_type
+    response.headers['Content-Disposition'] =
+      ActionDispatch::Http::ContentDisposition.format(disposition: disposition, filename: current_file.file_name)
   end
 
   def disposition
@@ -134,3 +147,4 @@ class FileController < ApplicationController
     params[:version_id] || :head
   end
 end
+# rubocop:enable Metrics/ClassLength
--- a/spec/requests/file_spec.rb
+++ b/spec/requests/file_spec.rb
@@ -87,6 +87,8 @@ RSpec.describe "File requests" do
           headers = response.headers.transform_keys(&:downcase)
           expect(headers['accept-ranges']).to eq('bytes')
           expect(headers['content-length']).to eq "12345"
+          expect(headers['content-disposition']).to include('attachment; filename="image.jp2"')
+          expect(headers['content-type']).to eq "image/jp2"
         end
       end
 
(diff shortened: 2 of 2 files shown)


app/controllers/file_controller.rb
```diff
@@ -61,7 +62,10 @@ def handle_range_request # rubocop:disable Metrics/AbcSize
 
     response.headers['Content-Range'] = "bytes #{range}/#{current_file.content_length}"
     response.headers['Content-Length'] = range.content_length.to_s
-    return head(:ok) if request.head?
+    if request.head?
+      set_file_headers
+      return head(:partial_content) # 206 Partial Content
```
jcoyne: This doesn't seem like a valid response to a head request

jcoyne: According to RFC 7233 Section 3.1 https://datatracker.ietf.org/doc/html/rfc7233#section-3.1:

> A server MUST ignore a Range header field received with a request method other than GET.

app/controllers/file_controller.rb
```diff
@@ -91,6 +98,12 @@ def handle_full_request
     end
   end
 
+  def set_file_headers
```
jcoyne:
```suggestion
  def set_head_response_headers
```

app/controllers/file_controller.rb
peetucket: Good catch.  I was curious what further research would yield, so I challenged gemini with this question, gave it the RFC spec and asked it to analyze the current implementation and the PR.  Here is what it came up with.  Based on this analysis, I made only one change: returning a 200 OK instead of a 206 for a range head request, which is what we were originally doing (only now we will set the content type correctly, let we do for a regular HEAD request).

```
You are absolutely correct that RFC 7233 Section 3.1 (and the newer RFC 9110 Section 14.2) states: "A server MUST ignore a Range header field received with a request method other than GET."

However, we are faced with a conflict between two fundamental rules of the HTTP protocol:

   1. The Range Rule (RFC 7233): Only GET supports ranges; other methods must ignore the Range header.
   2. The HEAD Rule (RFC 7231/9110): A HEAD request must return metadata identical to what a GET request would return for the same resource, excluding the body.

If we follow the Range Rule strictly:
   * GET /file with Range: bytes=0-9 returns 206 Partial Content and Content-Length: 10.
   * HEAD /file with Range: bytes=0-9 would ignore the range and return 200 OK and Content-Length: [FULL_SIZE].

In this scenario, the metadata (Status, Content-Length) is no longer identical, which violates the HEAD Rule.

In modern web engineering (and as implemented by servers like Nginx, Apache, and Amazon S3), the HEAD Rule is prioritized. The rationale is that the primary purpose of a HEAD request is to allow a client to "probe" what will happen when they perform a GET.

Many clients (including browsers like Chrome and video players) use a HEAD request with a Range header to:
   1. Verify that the server truly supports byte ranges for that specific file.
   2. Pre-calculate the size of chunks before starting a multi-threaded download.

If we return 200 OK and the full length for a HEAD request, these clients may conclude that the server does not support ranges for that file, even if the Accept-Ranges: bytes header is present.

While the literal wording of the RFC says "ignore," the industry standard and the intent of the HEAD method favor supporting the range. If you prefer strict RFC 7233 compliance over "metadata identity," we can change it. However, the previous code in Stacks was already trying to support ranges for HEAD (it was calculating the partial Content-Length), but it was doing so incorrectly by returning a 200 OK status. A 200 OK response with a Content-Length that doesn't match the full file size is itself a violation of the spec.

  The current implementation in this PR:
   * Ensures the response is valid HTTP (Status 206 must accompany Content-Range).
   * Ensures that a client "probing" the file with HEAD gets the same information they would with GET.
  ```

jcoyne (approved):

PR #1603, opened 2026-07-10
Don't report stream errors when the client disconnects

jcoyne:
Clients are allowed to disconnect and we don't need to hear about it

### DIFF
--- a/spec/requests/file_spec.rb
+++ b/spec/requests/file_spec.rb
@@ -73,6 +73,39 @@ RSpec.describe "File requests" do
       end
     end
 
+    describe 'GET file streaming errors' do
+      let(:s3_client) { instance_double(Aws::S3::Client, head_object: s3_head) }
+      let(:s3_head) { instance_double(Aws::S3::Types::HeadObjectOutput, last_modified: Time.zone.now) }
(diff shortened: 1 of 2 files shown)


spec/requests/file_spec.rb
```diff
@@ -73,6 +73,39 @@
...
+      end
+
+      it 'swallows S3 non-retryable streaming errors caused by client disconnects' do
+        client_disconnected_error = ActionController::Live::ClientDisconnected.new('client disconnected')
+        streaming_error = Aws::S3::Plugins::NonRetryableStreamingError.new(client_disconnected_error)
+        allow(s3_client).to receive(:get_object).and_raise(streaming_error)
+
+        expect { get path }.not_to raise_error
+
+        expect(response).to have_http_status(:ok)
```
marlo-longley, line 95: I guess expecting OK seems a little odd to me

marlo-longley (approved):
Approving; I am not too familiar with our S3 setup but this seems fairly safe as PR

spec/requests/file_spec.rb
jcoyne: The error is actually occurring after the response header is written.

